PR #9, opened 2018-05-10
Readme enhancements

killerboot:


README.md
```diff
@@ -23,165 +23,130 @@ Start Alfresco Content Services  using docker-compose or Kubernetes, containing:
...
-### Add the incubator helm repo:
+Check that the repository was added
 ```bash
-helm repo add alfresco-incubator http://kubernetes-charts.alfresco.com/incubator
 helm repo list
 ```
+#### Install the ingress service:
+Deploy nginx:
+```bash
+helm install stable/nginx-ingress --set rbac.create=true
```
svidrascu: Would be better if you pinpoint a version for the ingress here.

killerboot: That's a valid concern, but I think we agreed that it should work with latest (unless there is a bug in ingress).
@derekhulley Can you please confirm? Otherwise I can put the latest chart version that is currently "tested" (0.18.0).

derekhulley: The rate of change of the software means that the fixed version will fall behind virtually immediately.
We want our master branch to be current and therefore the instructions should reflect support for the latest versions of the software.  We might pin things once we have support branches.

derekhulley: It also goes without saying that we will review any failures that are highlighted as a result of this policy and fix our charts accordingly.

PR #12, opened 2018-05-16
REPO-3531: Limit JVM Memory in Kubernetes Pods 

amohammedalfresco:
- Follow best practices to limit JVM memory for Compose & Helm charts to guard against Java OOM crashes
- Define container memory and assign a proportion of it to JVM depending on ACS component
- Add comment to explain memory management for ACS containers

### DIFF
--- a/docker-compose/docker-compose.yml
+++ b/docker-compose/docker-compose.yml
@@ -33,7 +33,7 @@ services:
             - 8082:8080 #Browser port
 
     alfresco-pdf-renderer:
-        image: alfresco/alfresco-pdf-renderer:1.0
+        image: alfresco/alfresco-pdf-renderer:1.1
         mem_limit: 1g
         environment:
             JAVA_OPTS: " -Xms256m -Xmx256m"
--- a/docker-compose/docker-compose.yml
+++ b/docker-compose/docker-compose.yml
@@ -33,7 +33,7 @@ services:
             - 8082:8080 #Browser port
 
     alfresco-pdf-renderer:
-        image: alfresco/alfresco-pdf-renderer:0.10
+        image: alfresco/alfresco-pdf-renderer:1.0
         mem_limit: 1g
         environment:
             JAVA_OPTS: "$JAVA_OPTS -Xms256m -Xmx256m"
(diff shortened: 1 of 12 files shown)


docker-compose/docker-compose.yml
```diff
@@ -1,10 +1,19 @@
-version: "3"
+# This docker-compose file will spin up an ACS cluster on a local host or on a server and it requires a minimum of 16GB Memory to distribute among containers.
+# Limit container memory and assign X percentage to JVM.  There are couple of ways to allocate JVM Memory for ACS Containers
+# For example: 'JAVA_OPTS: "$JAVA_OPTS -XX:+PrintFlagsFinal -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap"'
+# But, as per Oracle docs (https://docs.oracle.com/javase/9/gctuning/parallel-collector1.htm#JSGCT-GUID-CAB83393-3438-44ED-98F0-D15641B43C7D)
+# If container memory is not explicitly set, then the above flags will default max heap to 1/4th of container's memory which may not be ideal. 
+# Hence, setting up explicit Container memory and then assigning a percentage of it to the JVM for performance tuning.
+
+# Using version 2 as 3 does not support resource constraint options (cpu_*, mem_* limits) for non swarm mode in Compose
```
montgolfiere: Do we no why this limitation was introduced (causing us to downgrade) ? Will this limit our ability to upgrade &/or make use of future docker-compose options ?

amohammedalfresco: Ref: https://docs.docker.com/compose/compose-file/#resources
```
Looking for options to set resources on non swarm mode containers?

The options described here are specific to the deploy key and swarm mode. If you want to set resource constraints on non swarm deployments, use Compose file format version 2 CPU, memory, and other resource options. If you have further questions, refer to the discussion on the GitHub issue docker/compose/4513.
```
As per the documentation and while using Version `3`, we were unable to set resource limits for non-swarm mode containers (i.e. Docker compose).

dcaruana: Am wondering if it's really a problem going with v2 - it works with latest docker releases, v1 is still accepted. We're not aiming for Docker swarm.

PR #24, opened 2018-06-20
Fix repository short name in values.yaml for search alfresco-host

amohammedalfresco:
- Use alfresco-cs short name in search host


alandavis:
Looks okay to me

PR #31, opened 2018-06-28
REPO-3425 tika docker transformer

alandavis:
Still need to generate images from helm-components.puml files


helm/alfresco-content-services/values.yaml
```diff
@@ -146,6 +146,39 @@ libreoffice:
     maxTransforms: 99999
     maxTransformSeconds: 1800
 
+# Declares the alfresco-libreoffice service used by the content repository
```
montgolfiere: Trivial: Declares the alfresco-tika service ...

docker-compose/docker-compose.yml
```diff
@@ -34,29 +36,37 @@ services:
             - 8082:8080 #Browser port
 
     alfresco-pdf-renderer:
-        image: alfresco/alfresco-pdf-renderer:1.1
+        image: alfresco/alfresco-pdf-renderer:1.2
```
montgolfiere: Why did the other 3 receive a version bump as part of this PR / set of commits ?

I guess I should review the PRs/commits in main code base first (and then packaging / deployment) ;-)

alandavis: The PO for the repo team wanted all the transformers lock stepped to together to have the same version, so when one is released they all are. Originally they all had separate release cycles. This may not be the case in the future.

helm/alfresco-content-services/values.yaml
alandavis: fixed

PR #41, opened 2018-07-10
REPO-3574: K8s and HELM: reference or basis of production

derekhulley:
Gathering feedback from ToK.


derekhulley:
We would like to add language to ensure that customers understand that, out of the box, the helm charts act as a reference deployment.  Please submit your comments and change suggestions.

AndyHunt66:
I have no problem with this specific change, but this is not the only page that gives the impression that the charts are a production-ready solution

derekhulley:
We have an open DOC issue and there is nothing in here that I can see that implies it is not a reference.

PR #28, opened 2018-06-26
SEARCH-825 (Create a helm chart for Alfresco Zeppelin)

tunaaksoy:

PR #42, opened 2018-07-10
SEARCH-825 (Create a helm chart for Alfresco Zeppelin) (#28)

tunaaksoy:
SEARCH-825 (Create a helm chart for Alfresco Zeppelin)

### DIFF
--- a/helm/alfresco-content-services/requirements.yaml
+++ b/helm/alfresco-content-services/requirements.yaml
@@ -10,7 +10,7 @@ dependencies:
   repository: https://kubernetes-charts.storage.googleapis.com/
   condition: postgresql.enabled
 - name: alfresco-search
-  version: 0.0.3
+  version: 0.0.4
   repository: http://kubernetes-charts.alfresco.com/stable
   condition: alfresco-search.enabled
 - name: alfresco-infrastructure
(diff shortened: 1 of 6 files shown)


tunaaksoy:
Is the release process described somewhere? Next time I can do the release if you want.

I think all you need to do is to merge this PR to support/HF/1.0, update Chart.yaml and helm-chart-releases.md files. Once the build on support/HF/1.0 is green then release it from there. Not so sure if I missed any steps.

killerboot:
I noticed that insight engine is not started up by default. Is it intentional?
The tests on a branch will not pass as the number of pods was wrong. I corrected the number back to 10, should pass now.

tunaaksoy:
Yes, that's intentional. I did not want to change the existing behaviour. For you nothing will change. You have the option to swap Search Services with Insight Engine if you want but you need to specify that (--set alfresco-search.type="insight-engine" --set alfresco-search.registryPullSecrets="quay-registry-secret"). Also Zeppelin has be explicitly enabled (--set alfresco-insight-zeppelin.enabled="true" --set alfresco-search.alfresco-insight-zeppelin.registryPullSecrets="quay-registry-secret") otherwise it won't be available.

tunaaksoy:
Thanks for fixing the number of pods. I was not so sure if the number of pods was correct or not.

tunaaksoy:
The bamboo build is green.

PR #49, opened 2018-07-11
Feature/repo 3545 nginx file limits

Epurashu:


helm/alfresco-content-services/values.yaml
```diff
@@ -223,7 +229,7 @@ alfresco-search:
 
 # choose if you want network policy enabled
 networkpolicysetting:
-  enabled: true
+  enabled: false
```
amohammedalfresco: @Epurashu: We want the `networkpolicysetting` to be enabled by default.

nicdoye:
A bit late, but is there any way to create the 1MB file without embedding the whole thing in the JSON?https://github.com/Alfresco/acs-deployment/pull/49/files#diff-4e7abf9fddeaee75ee9f045d488bfa0a

It really messes up running `git diff` on the command line. 😺

PR #52, opened 2018-07-18
Clarify Kops configuration

skopf:
- added hint to use EKS instead of Kops
- added note about configuring Kops after getting feedback

### DIFF
--- a/docs/helm-deployment-aws_cloud.md
+++ b/docs/helm-deployment-aws_cloud.md
@@ -22,7 +22,7 @@ To run the Alfresco Content Services (ACS) deployment on AWS in a Kops provided 
 
 Kops requires certain configurations on your AWS account. Please follow the guide
 [here](https://github.com/kubernetes/kops/blob/master/docs/aws.md#setup-your-environment)
-through "Creating your first cluster". We will create the cluster for ACS in the next step.
+through "Creating your first cluster". The cluster for ACS will be created in the next step.
 
 ### Setting up Kubernetes cluster on AWS with Kops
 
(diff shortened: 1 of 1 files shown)


docs/helm-deployment-aws_cloud.md
```diff
@@ -16,6 +18,11 @@ To run the Alfresco Content Services (ACS) deployment on AWS with Kubernetes req
 
 **Note:** You don't need to clone this repository to deploy Alfresco Content Services.
 
+### Prepare and configure Kops
+
+Kops requires certain configurations on your AWS account. Please follow the guide
+[here](https://github.com/kubernetes/kops/blob/master/docs/aws.md#setup-your-environment)
+through "Creating your first cluster". We will create the cluster for ACS in the next step.
```
killerboot: Probably a better wording would be "The cluster for ACS will be created in the next step".

docs/helm-deployment-aws_cloud.md
```diff
@@ -1,9 +1,11 @@
 # Alfresco Content Services Deployment with Helm on AWS
 
+**Hint:** Consider [deploying ACS into EKS](./helm-deployment-aws_eks.md) instead of setting
```
killerboot: Probably better to add a note to the links in https://github.com/Alfresco/acs-deployment/blob/master/README.md#deployment-options that EKS is the recommended way for AWS deployments.

skopf:
@anxumalo is already on the review list.

docs/helm-deployment-aws_cloud.md
skopf: 1. This note is for people who land here directly, e.g. by deep link from another project
2. Are we sure that EKS is now the recommended way?

PR #53, opened 2018-07-19
Added section about custom infrastructures

skopf:
We should make clear that these charts are not designed to be used as-is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Alfresco Content Services deployed via `docker-compose` or Kubernetes contains t
 * [Customizing your deployment](docs/customising-deployment.md)
 
 ## Production environments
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,6 +35,15 @@ Alfresco Content Services deployed via `docker-compose` or Kubernetes contains t
 * [Deploying using Docker Compose](docs/docker-compose-deployment.md)
 * [Customizing your deployment](docs/customising-deployment.md)
 
+## Production environments
+The HELm charts in this repository provide a PostgreSQL database in a docker container and do not configure
```
killerboot: A small typo here "HELm"

derekhulley: We agreed, Alex, that small typos would just be done directly rather than holding up the review.

PR #62, opened 2018-08-15
REPO-3721: Admin console link

Epurashu:
docker-compose : Add external alfresco.host and alfresco.port to JAVA_OPTS for repository image in order to fix AOS
   docker-compose : Add external aflresco.host alfresco.port alfresco.context and alfresco.protocol to JAVA_OPTS for share image in order to fix broken admin console link
   helm : Add CATALINA_OPTS to config-share. Add alfresco.proxy value to fix broken admin console link

### DIFF
--- a/docker-compose/docker-compose.yml
+++ b/docker-compose/docker-compose.yml
@@ -78,7 +78,7 @@ services:
             - 8099:8099
 
     share:
-        image: alfresco-share-test
+        image: alfresco/alfresco-share:6.0
         mem_limit: 1g
         environment:
             REPO_HOST: "alfresco"
--- a/docker-compose/docker-compose.yml
+++ b/docker-compose/docker-compose.yml
@@ -29,6 +29,8 @@ services:
                 -Dtika.url=http://tika:8090/
                 -Dsfs.url=http://shared-file-store:8099/
                 -Dshare.host=localhost
+                -Dalfresco.host=localhost
+                -Dalfresco.port=8082
                 -Ddeployment.method=DOCKER_COMPOSE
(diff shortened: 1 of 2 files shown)


docker-compose/docker-compose.yml
```diff
@@ -76,12 +78,19 @@ services:
             - 8099:8099
 
     share:
-        image: alfresco/alfresco-share:6.0
+        image: alfresco-share-test
```
alandavis: Looks like we are going to have to wait for a Share release.
Might be better to not change the Share version and to get the rest of the changes merged to master so they will be picked up when there is a Share release.

docker-compose/docker-compose.yml
```diff
@@ -76,12 +78,19 @@ services:
...
-            - "CATALINA_OPTS= -Xms500m -Xmx500m"
+            REPO_HOST: "alfresco"
+            REPO_PORT: "8080"
+            JAVA_OPTS: "
+                -Xms500m
+                -Xmx500m
+                -Dalfresco.host=localhost
+                -Dalfresco.port=8082
+                -Dalfresco.context=alfresco
+                -Dalfresco.protocol=http
```
alandavis: Is this correct? Would it normally be https externally?

Epurashu: We are currently not using HTTPS for our docker-compose deployment.

docker-compose/docker-compose.yml
Epurashu: This one slipped into the commit, I`m going to revert this change.

PR #63, opened 2018-08-16
REPO-3704: s3 compatibility for the kubernetes helm chart.

svidrascu:


docs/helm-deployment-aws_cloud.md
```diff
@@ -402,6 +402,30 @@ helm install alfresco-incubator/alfresco-content-services \
 
 **Note:** When using an external database, make sure that the database is reachable by the K8S nodes and is up and ready.
 
+### Using the Alfresco S3 Connector
+
+If you have the S3 connector amp applyed to your docker image then you could enable it in kubernetes by setting the following values:
```
derekhulley: "applied"

killerboot:
Another question, should we make EFS and S3 mutually exclusive in the config?

svidrascu:
Yes we have automated testing in mind, however i do not think the tests will reach this repo.

svidrascu:
Considering that efs roughly costs 10 times more than s3 i would believe they should be mutually exclusive. Even with cross region replication it is still cheaper.

docs/helm-deployment-aws_cloud.md
svidrascu: fixed, thanks 👍

helm/alfresco-content-services/README.md
```diff
@@ -81,3 +81,16 @@ Parameter | Description | Default
 `alfresco-search.ingress.enabled` | Enable external access for Alfresco Search Services | `false`
 `alfresco-search.ingress.basicAuth` | If `alfresco-search.ingress.enabled` is `true`, user need to provide a `base64` encoded `htpasswd` format user name & password (ex: `echo -n "$(htpasswd -nbm solradmin somepassword)"` where `solradmin` is username and `somepassword` is the password) | None
 `alfresco-search.ingress.whitelist_ips` | If `alfresco-search.ingress.enabled` is `true`, user can restrict `/solr` to a list of IP addresses of CIDR notation | `0.0.0.0/0`
+`s3connector.enabled` | Switch this to true if you have access to the s3connector Amp | `false`
+`s3connector.config` | S3 Configuration https://docs.alfresco.com/s3connector/references/s3-contentstore-ref-config-props.html. Example: s3connector.config.bucketName: myS3Bucket | `{}`
+`s3connector.secrets` | S3 Secrets Configuration https://docs.alfresco.com/s3connector/references/s3-contentstore-ref-config-props.html. Example: s3connector.secrets.accessKey: AJJJJJJJJ | `{}`
+
```
killerboot: Please remove white lines.

docs/helm-deployment-aws_cloud.md
```diff
@@ -402,6 +402,31 @@ helm install alfresco-incubator/alfresco-content-services \
...
+
+If you have the S3 connector amp applied to your docker image then you could enable it in kubernetes by setting the following values:
+
+```
+helm install alfresco-incubator/alfresco-content-services \
+--set externalProtocol="https" \
+--set externalHost="$EXTERNALHOST" \
+--set externalPort="443" \
+--set repository.adminPassword="$ALF_ADMIN_PWD" \
+--set alfresco-infrastructure.persistence.efs.enabled=true \
```
killerboot: I suspect we are leaving EFS for DB and Solr, but those are not clustered. Increasing the number of desired pods will not make them clustered as well. Should EFS be removed in this example and default EBS will be used?

helm/alfresco-content-services/values.yaml
```diff
@@ -253,6 +253,7 @@ share:
 persistence:
   existingClaim: "alfresco-volume-claim"
   repository:
+    enabled: true
```
killerboot: I think this switch needs to be documented together with `s3connector.enabled` describing what they do.
Please take a look at the table of possible values for the Helm chart here:
https://github.com/Alfresco/acs-deployment/tree/master/helm/alfresco-content-services#configuration

svidrascu: done

docs/helm-deployment-aws_cloud.md
svidrascu: S3 is not a default so using EFS still stands for clustering / DB and Solr.

svidrascu: Yes, Alex i would still leave it like this, with EFS, since EBS has a lot of issues moving from one node to another in case one of the pods gets rescheduled. But you could without issues do 
 --set alfresco-infrastructure.enabled=false to go for the default gp2.

PR #65, opened 2018-08-17
MNT-19918: Remove force ssl

svidrascu:


killerboot:
Please increment the version of the chart as well

svidrascu:
1.0.3 was not released
So I will do the chart version bump after releasing 1.0.3

PR #75, opened 2018-08-31
repo-3802 Add nginx config for blocking alfresco/s/api/solr

mmuller88:


skopf:
LGTM. However, if someone deploys our HELM charts with an ingress-controller other than nginx, those rules won't be applied.

PR #77, opened 2018-09-03
REPO-3802 Unprotected access to Solr APIs

mmuller88:
- forbid more urls
- add postman tests


skopf:
After looking at the actual URLs we block for Share, it became clear that the proxy service in Share allows to control everything from `/api/...` onwards.
The additional URLs we are now blocking are all different prefixes for `/api/...`.

PR #78, opened 2018-09-10
ATS-92 : The first Deployment update for ACS (Repo) to use ATS

DenisGabriela:


docker-compose/docker-compose.yml
```diff
@@ -42,44 +42,61 @@ services:
         ports:
             - 8082:8080 #Browser port
 
+    router:
```
derekhulley: The name "router" is not specific enough.

docker-compose/docker-compose.yml
```diff
@@ -42,44 +42,61 @@ services:
...
         mem_limit: 1g
         environment:
             JAVA_OPTS: " -Xms256m -Xmx256m"
+            FILE_STORE_URL: "http://shared-file-store:8099/alfresco/api/-default-/private/sfs/versions/1/file"
         ports:
             - 8093:8090
 
     shared-file-store:
-        image: alfresco/alfresco-shared-file-store:0.1
+        image: quay.io/alfresco/alfresco-shared-file-store:ATS-94-1.0-SNAPSHOT
```
derekhulley: Never snapshots.  Please use a released version.

docker-compose/docker-compose.yml
```diff
@@ -42,44 +42,61 @@ services:
...
         environment:
             JAVA_OPTS: " -Xms256m -Xmx256m"
         ports:
             - 8099:8099
+        volumes:
+            - shared-file-store-volume:/tmp/Alfresco
```
derekhulley: Is this specific enough?  Could we use "/tmp/Alfresco/sfs"?

helm/alfresco-content-services/values.yaml
```diff
@@ -190,8 +210,8 @@ tika:
 filestore:
   replicaCount: 1
   image:
-    repository: alfresco/alfresco-shared-file-store
-    tag: "0.1"
+    repository: quay.io/alfresco/alfresco-shared-file-store
+    tag: "ATS-94-1.0-SNAPSHOT"
```
derekhulley: No snapshots, please.

montgolfiere: Yes, agreed. This is known to be WIP (for early review) until we resolve REPO-3749 (and ATS-94).

montgolfiere:
@killerboot good point. We need a decision @pnirwal that we can deploy T-Router image to dockerhub in order to support docker-compose option for ATS with ACS ?

derekhulley:
There should be a few readme updates to reflect the new contents, please.

helm/alfresco-content-services/values.yaml
derekhulley: Publish images to Docker Hub as -EA e.g. ATS-94-1.0-EA1 and then we can use those.
Since the code is in the Enterprise domain, and the images are EA, we can pull them or replace them as required later.

DenisGabriela:
Thanks for all your feedback. I will sure add details on the readme, also will address the other comments.

docker-compose/docker-compose.yml
```diff
@@ -42,44 +42,61 @@ services:
...
+          IMAGEMAGICK_URL: "http://imagemagick:8090"
+          PDF_RENDERER_URL : "http://alfresco-pdf-renderer:8090"
+          LIBREOFFICE_URL : "http://libreoffice:8090"
+          TIKA_URL : "http://tika:8090"
+        links:
+          - activemq
+
     alfresco-pdf-renderer:
-        image: alfresco/alfresco-pdf-renderer:1.3
+        image: alfresco/alfresco-pdf-renderer:2.0.0-RC2
```
derekhulley: Check with @montgolfiere and @pnirwal , but these images should not be public as agreed with management, yesterday.  This compose requires quay.io, anyway.  The original 1.N images can remain in Docker Hub as they are required for the current Enterprise trial.

helm/alfresco-content-services/values.yaml
```diff
@@ -53,13 +53,33 @@ apiexplorer:
...
+    initialDelaySeconds: 140
+    periodSeconds: 120
+    timeoutSeconds: 60
+
 # Declares the alfresco-pdf-renderer service used by the content repository
 # to transform pdf files
 pdfrenderer:
   replicaCount: 1
   image:
     repository: alfresco/alfresco-pdf-renderer
```
derekhulley: Management are expecting that the updated images remain in quay.io.
The 1.N line can stay in Docker Hub as they are used directly by the current trial, which does not required quay.io access.

DenisGabriela: Ok, we'll remove transformers 2.0.0-RC1, 2.0.0-RC2 and latest from Docker Hub.

montgolfiere: @pnirwal @derekhulley Hmm ... my understanding (not being present at "the meeting") was that this applies to Transform Router image only ? 

Otherwise, we end up with fork of maintenance process (build & release plan) for T-Engines (if we ever need to back-port hot-fixes from 2.N to 1.N).

derekhulley: That is, unfortunately, a detail.
The box drawn with JN yesterday went around all the images.

derekhulley: The decision as to whether "scale" is protected is what is important.

montgolfiere: OK. We're tracking via ATS-122. I've added the "detail" consequence as a [comment](https://issues.alfresco.com/jira/browse/ATS-122?focusedCommentId=584605&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-584605).

killerboot:
Can you please update the diagrams:
https://github.com/Alfresco/acs-deployment/tree/master/docs/diagrams

montgolfiere:
@killerboot Helm diagram makes sense. However, I'm not sure about Docker Compose diagram (see PRODDEL-779) unless we have a way to for single D/C file to switch Docker Registry (from dockerhub to quay.io) and not fail when using dockerhub.

killerboot:
Is it possible to add some basic tests via Postman at this moment or is it too early?
https://github.com/Alfresco/acs-deployment/tree/master/test/postman

montgolfiere:
I suggest too early (at least not part of original ATS-92 A/Cs). How much will it relate to ACS Renditions Test ? 

Otherwise we should backlog a new story to design & implement suitable sanity test (eg. not sure how Postman/Newman can be used to produce/consume to/from Qs).

killerboot:
I think we are close to finishing this one and merging it to master. We will deal with ActiveMQ helm chart later in [REPO-3682](https://issues.alfresco.com/jira/browse/REPO-3682) and tests in [REPO-3690](https://issues.alfresco.com/jira/browse/REPO-3690)

DenisGabriela:
I've updated the diagrams. Please take a look.

killerboot:
Green build:
https://bamboo.alfresco.com/bamboo/browse/PLAT-ACSHELM70-17

montgolfiere:
@derekhulley Can you please re-review and unblock the merge - thanks :-)

PR #82, opened 2018-09-18
REPO-3832: Update docs to recommend HTTPS

killerboot:
This is to address MNT-20007: "alfresco-docker-base-tomcat sets cookie secure
flag to true, requiring https for cookies to be transmitted"

### DIFF
--- a/docs/helm-deployment-aws_cloud.md
+++ b/docs/helm-deployment-aws_cloud.md
@@ -269,6 +269,8 @@ helm install stable/nginx-ingress \
 ```
 Adjust the above values accordingly (ex: `--version` of `nginx-ingress`, `controller.service.annotations` etc.)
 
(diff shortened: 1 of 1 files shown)


docs/helm-deployment-aws_cloud.md
```diff
@@ -269,6 +269,8 @@ helm install stable/nginx-ingress \
 ```
 Adjust the above values accordingly (ex: `--version` of `nginx-ingress`, `controller.service.annotations` etc.)
 
+Please note that ACS Docker image is running Tomcat in secure mode (see the [Dockerfile](https://github.com/Alfresco/alfresco-docker-base-tomcat/blob/master/Dockerfile)). It implies that Tomcat will only send cookies via HTTPS connection. **Deploying the ingress without HTTPS (Option 1) is not recommended,** as some of the functionality will not work.
```
derekhulley: Probably worth changing the actual "Option 1" to "Option 1 (not recommended)".  Not sure why it would be option 1, then :-)

PR #93, opened 2018-10-17
ACA-1847: integrate ADW image and app proxy

DenysVuika:
## Details

The PR adds extra Alfresco Digital Workspace 1.0.0 from quay.io.
The resulting application is served at port 8085.

In order to solve the problems with the CORS the application comes with the proxy container at port 8084 that allows redirecting requests to the underlying repository.

See also: https://issues.alfresco.com/jira/browse/ACA-1847

## Result
`http://localhost:8084` points to the ADW instance
`http://localhost:8084/alfresco` proxies the calls from the app to Repository without CORS issues

## Testing

```sh
docker-compose up
open http://localhost:8084
```

Use default `admin:admin` to log into the app.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ Alfresco Content Services deployed via `docker-compose` or Kubernetes contains t
 3. A Postgres DB  
 4. Alfresco Search Services (Solr6)
 5. Alfresco Transform Service
+6. Alfresco Digital Workspace
 
 ## Deployment options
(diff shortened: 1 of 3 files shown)


derekhulley:
Also, please add details to the CHANGELOG and appropriate documentation updates.

DenysVuika:
Ok, thanks for feedback. I will do that as part of the https://github.com/Alfresco/acs-deployment/pull/94

PR #95, opened 2018-10-19
WIP: REPO-3867: AWS Load Tests: Integrate yourkit profiling agent into the docker images for ACS Repository

aforascu:
- added a new volume to docker-compose for the YourKit agent snapshots that can be taken on ACS deployment
- the commented YourKit startup parameters can be added as CATALINA_OPTS or JAVA_OPTS on alfresco service
- a new nginx pod has been created expose and allow read access for the user on the ACS Snapshots taken by the agent


gavincornwell:
This was created for a block of work that never happened and is no longer planned so closing this PR.

PR #102, opened 2018-11-09
Deploy 563: Stable Infrastructure Chart for ACS 6.1 (with ActiveMQ)

cristina-sirbu:
DBP chart brings its own Alfresco Infrastructure. So when we bring in the ACS chart we disable the requirement to bring another Infrastructure Chart.
We need a default for message broker URL property to pick up the ActiveMq service that comes with DBP's Alfresco Infrastructure.


derekhulley:
Is there an equivalent change for acs-community-deployment?

nicdoye:
lgtm-sheep git

PR #106, opened 2018-11-12
SEARCH-1253: Create tests for /search api end points for acs-deployment

meenalbhave:
Added 2 simple tests for search end point in acs test collection. Would complete the task after the initial review.


paulbrodner:
@killerboot 
* I've fixed the tests above, [PLAT-ACSHELM94-PDHC-3](https://bamboo.alfresco.com/bamboo/browse/PLAT-ACSHELM94-PDHC-3) passing for this Pull Request

* I've also updated the tests and introduced "protocol" variable, so we can reuse the tests on "http" not only "https".  
  * For this to work I've added the protocol global variable in the [Run Postman Collection task](https://bamboo.alfresco.com/bamboo/build/admin/edit/editBuildTasks.action?buildKey=PLAT-ACSHELM-DI) `(...) --global-var "protocol=https" (...)`
  * Master seems to work (no regression introduce) [PLAT-ACSHELM-303](https://bamboo.alfresco.com/bamboo/browse/PLAT-ACSHELM-303)

What do you think - should we go with this approach or should I revert the change (regarding protocols) 

This PL is still a WIP - just talked with @meenalbhave - we will need to add a couple of test under `acs-validate-volume-collection.json`

killerboot:
@paulbrodner Having a configurable protocol won't hurt, but all the tests are running against environment with https. One of the reasons is that Tomcat running in secure mode is not sending the cookies over http, so this setup is not really a supported one. Running the helm chart on local machine with http to run the tests against it seems less relevant as memory consumption is a bit high.
Personally I don't see much value, but, if you still want this change, could you please separate it into a different PR, as this one doesn't even show the diff as it is too large.

Tests wise, if search team is happy with the coverage, I'm ok as well. Only one concern is the test names, especially "search-name" is not descriptive for a test, please change those.

meenalbhave:
@killerboot, @paulbrodner, with regards the test name, search-using-name and search-using-cmis may be?

derekhulley:
S&D discovery has charts that should be tested, already.
This project should just validate enough to ensure that the introduction of the search charts is done.

paulbrodner:
Thank you @killerboot  (_like this username 👍_ ) 
* changes regarding "protocol" were removed from this PR and introduced into a separate PR: #107 (thank you for accepting that)
* renamed search tests:
  * search-name as search-api-validation
  * search-cmis as search-cmis-validation

Regarding 
> This project should just validate enough to ensure that the introduction of the search charts is done.

I believe we are ok with this - i've double checked with @meenalbhave  today!

**Build Validation:**
* Build Plan is passing: [PLAT-ACSHELM94-4](https://bamboo.alfresco.com/bamboo/browse/PLAT-ACSHELM94-4) - if this is OK with you @killerboot  - I can merge it to master!

**Note:**
It is a pain that we cannot diff the previous changed on `acs-test-collection.json` - it seems this is due to a test that is validating file-size bigger than 1mb

PR #114, opened 2018-11-20
Fix AOS base URL

killerboot:


derekhulley:
Same required for ace-community-deployment or do we stick with the root for it as adw is not interfering there?
@skopf

killerboot:
@derekhulley The protocol/host/port will be set according to the usual values we have already set in a-g.p or via env variables, regardless if there is a proxy or not.
This property controls the links to folders/documents in a _browser_ view of AOS, like https://myhost/alfresco/aos

The ADW being mapped to root context needs to be fixed separately.

PR #120, opened 2018-11-27
WIP: Experiment: SEARCH-1305: speedup packaging of Helm chart on dev environment

paulbrodner:
**Features:**
within "_helm/alfresco-content-services_" folder:
* ability to bump up the new release version automatically: `make tag` (and also push it to repository if we want `make release-tag`)
* ability to build or package the chart: `make package`
* ability to build or package the chart, not skipping the local helm repository: `make local-repo package`

**Note:**
You can mix the commands if you like: `make tag release-tag package` (this should auto-increment the version, commit the tag and push the updated Chart.yaml to repo

### DIFF
--- a/helm/alfresco-content-services/.cmd/new-version.sh
+++ b/helm/alfresco-content-services/.cmd/new-version.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+if [ $# -lt 1 ]; then
+   echo "Please pass the current version as first parameter so I can generate the new version!" 
+   echo "Example: $ $0 0.0.2 #will generate 0.0.3" 
+   exit 1 
+fi
+set -e
+echo $1 | awk -F. -v OFS=. 'NF==1{print ++$NF}; NF>1{if(length($NF+1)>length($NF))$(NF-1)++; $NF=sprintf("%0*d", length($NF), ($NF+1)%(10^length($NF))); print}'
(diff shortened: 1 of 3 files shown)


derekhulley:
This looks interesting.
Can you point to the equivalent changes in https://github.com/Alfresco/alfresco-search-deployment?

tunaaksoy:
The equivalent changes on our side can be found here: https://github.com/Alfresco/alfresco-search-deployment/pull/10

This is currently in review/verify state on our end.

paulbrodner:
@derekhulley as an overview, checkout this https://issues.alfresco.com/jira/browse/SEARCH-1305

derekhulley:
OK.  Now the change makes sense.
Note that the requirement to have a branch, a green build and a merge means that you can't update the _requirements_ during the release process.  We could relax some requirements but a green build on master should be considered a minimum (obviously) before charts are released.

tunaaksoy:
My concern is that we have the same files (new-version.sh and Makefile) in both projects and to maintain them will be difficult. Make file in this repo has one more target for starting up a local helm repo but there is no other change in the files. And if we think of other helm charts which might become a requirement for ACS helm chart will we apply the same pattern for them as well? It will be optional I think but there is a chance to use the files there as well which will make maintenance a problem. Not so sure if someone fixes an issue in the files in this repo will also make sure to fix the same issue in our chart (and probably others).

We also need to be careful when bumping up version numbers. It's OK locally but it's very easy to push that change as part of a PR/MR. You are not supposed to bump up the version every time when you commit. There is no hard rules though as far as I am aware of.

killerboot:
I'll close this PR, but leave the branch if you still need it.

PR #122, opened 2018-11-27
Update chart version after the release

tunaaksoy:


tunaaksoy:
It looks like a release has been done (v 1.1.6) and the chart version has not been updated to v1.1.7. Also another build was run which pushed v 1.1.6 to incubator again. I think we should come up with a better/easier release workflow for helm charts.

tunaaksoy:
Green build: https://bamboo.alfresco.com/bamboo/browse/PLAT-ACSHELM113-2

PR #125, opened 2018-12-04
ACE-5900: move ADW to the "/workspace" route

DenysVuika:
As per requests:

* move ADW to the "/workspace" root (see https://issues.alfresco.com/jira/browse/ACE-5900)
* expose ACS via the "/"
* remove old comments from the docker-compose


DenysVuika:
Thanks, feel free to merge it then if there are no concerns.

gparapadakis:
Can I please check (for clarity, as we may need to adjust the trials documentation). Does this fix mean that "/" will no longer default to ADW, in RC2?

DenysVuika:
Yes, that’s correct. You can check ACE-5900 for the details and discussion in comments

PR #126, opened 2018-12-05
ATS-215 : Update Xmx memory values for the Transform Service referenc…

tzclucian:
…e deployments

   - updated values


derekhulley:
Assuming it's been checked locally on dev laptops as well as build env.

montgolfiere:
Do we have guidance for minimium memory on a dev laptop ? I'm assuming at least 16Gb ?

This noted in instructions for minikube (but not specifically for docker-compose).

PR #128, opened 2018-12-05
WIP: REPO-4091 aos

mmuller88:
* allow access to root which is needed from AOS
* add Postman tests for AOS


mmuller88:
We would need to have RC2 release before I merge can merge that. But please review the important changes.

PR #135, opened 2018-12-12
ATS-216: Update helm chart reference deployment (on k8s) with 2 replicas

CezarLeahu:
- update replicaCount from 1 to 2 for T-Router and T-Engines
- keep only 1 replica for Shared File Store (in case EFS is not enabled)
- update pod scheduling for different Availability Zones and K8s nodes

### DIFF
--- a/helm/alfresco-content-services/templates/deployment-filestore.yaml
+++ b/helm/alfresco-content-services/templates/deployment-filestore.yaml
@@ -24,6 +24,27 @@ spec:
       imagePullSecrets:
         - name: {{ .Values.registryPullSecrets }}
       {{- end }}
+      affinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 10
+            podAffinityTerm:
+              labelSelector:
+                matchExpressions:
+                - key: app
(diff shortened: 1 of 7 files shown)


montgolfiere:
FYI: we may need someone from Repo team to help us change the build scripts in order to get a green build please ! 

I assume this requires change in this area:

https://git.alfresco.com/Repository/acs-k8s-cluster/blob/master/scripts/test_acs_setup.sh#L25-29

... hence cc: @killerboot @mmuller88 @cristianturlica @alandavis as recent comitters to that project ;-)

Unfortunately, build scripts are separate from this repo so we may need to coordinate merges (across active branches, ie. master and support/HF/1.1) or else make build scripts sensitive to the branch ?

killerboot:
The check for number of pods is configured in the build plan. Last time it was 15, I suspect it is now 20.
I'll update the value for your branch and see how it goes.

killerboot:
green build:
https://bamboo.alfresco.com/bamboo/browse/PLAT-ACSHELM122-2

montgolfiere:
Ah, good to know - I hadn't reverse-engineered that far ;-)

PR #152, opened 2019-01-29
Global ACS Ingress configuration Helm values

igdianov:
This PR provides support for global ACS ingress configuration via Helm values and includes:

* fix: configure global values for ingress hostName, tls, secret, annotations 
* fix: replace maxUploadSize with 5000m (5g does not work in EKS)

### DIFF
--- a/helm/alfresco-content-services/templates/ingress-repository.yaml
+++ b/helm/alfresco-content-services/templates/ingress-repository.yaml
@@ -15,16 +15,33 @@ metadata:
       location ~ ^(/.*/s/api/solr/.*)$ {return 403;}
       location ~ ^(/.*/wcservice/api/solr/.*)$ {return 403;}
       location ~ ^(/.*/wcs/api/solr/.*)$ {return 403;}
+    # Global ingress annotations to be combined with specific ingress annotations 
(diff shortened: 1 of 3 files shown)


igdianov:
@svidrascu I am proposing to use global values to be able to propagate ACS ingress settings into embedded subcharts, i.e. DW, Solr, and any other add-on chart that may want to apply ACS global ingress settings as part of the Helm release.

We have been using this approach in Activiti 7 Helm chart.

svidrascu:
Ok, sounds good, but this means you also need to update the readmes related to the chart installations since global values are not propagated correctly over child charts trough --set option (caused by bug -> https://github.com/helm/helm/issues/4767 ), which we use all over the place in acs.

https://github.com/Alfresco/acs-deployment/blob/master/docs/helm-deployment-aws_kops.md
https://github.com/Alfresco/acs-deployment/blob/master/docs/helm-deployment-minikube.md#option-1-deploying-alfresco-content-services-from-the-chart-repository

killerboot:
This is solved by DEPLOY-655, see https://github.com/Alfresco/acs-deployment/pull/183

PR #158, opened 2019-02-14
REPO-4145 : Helm Chart does not start if alfresco-search is disabled

Epurashu:
Add external host and port for search in values.yaml
   If alfresco-search chart is not enabled, add if-else-end clause to config-repository.yaml to skip adding solr.host and solr.port from alfresco-search charts (null value)
   If alfresco-search chart is not enabled, add if-else-end clause to config-repository.yaml to add external values (if uncommented)
   If alfresco-search chart is not enabled, disable ingress trafic restriction to */solr/*
   If alfresco-search chart is not enabled, disable search network policy
   If alfresco-search chart is not enabled, disable trafic from in-cluster search pod.

### DIFF
--- a/helm/alfresco-content-services/templates/Rule_04-network-policy-repository.yaml
+++ b/helm/alfresco-content-services/templates/Rule_04-network-policy-repository.yaml
@@ -18,10 +18,12 @@ spec:
           matchLabels:
             app: nginx-ingress
 
+      {{- if index .Values "alfresco-search" "enabled" }}
       # Allow traffic from search
       - podSelector:
           matchLabels:
-            app: {{ template "alfresco-search.host" . }}
+            app:  {{ template "alfresco-search.host" . }}
+      {{ end }}
 
       # Allow traffic from share
       - podSelector:
@@ -70,11 +72,13 @@ spec:
       - podSelector:
           matchLabels:
              app: {{ template "content-services.shortname" . }}-repository
-      
+
+      {{ if index .Values "alfresco-search" "enabled" }}
       # Search (solr)
       - podSelector:
           matchLabels:
-             app: {{ template "alfresco-search.host" . }}
+             app:  {{ template "alfresco-search.host" . }}
+      {{ end }}
 
       # Share
       - podSelector:
--- a/helm/alfresco-content-services/templates/Rule_04-network-policy-repository.yaml
+++ b/helm/alfresco-content-services/templates/Rule_04-network-policy-repository.yaml
@@ -23,7 +23,7 @@ spec:
       - podSelector:
           matchLabels:
             app:  {{ template "alfresco-search.host" . }}
-      {{ end }}
+      {{- end }}
 
       # Allow traffic from share
       - podSelector:
@@ -73,12 +73,12 @@ spec:
           matchLabels:
              app: {{ template "content-services.shortname" . }}-repository
 
-      {{ if index .Values "alfresco-search" "enabled" }}
+      {{- if index .Values "alfresco-search" "enabled" }}
       # Search (solr)
       - podSelector:
           matchLabels:
              app:  {{ template "alfresco-search.host" . }}
-      {{ end }}
+      {{- end }}
 
       # Share
       - podSelector:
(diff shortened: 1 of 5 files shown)


helm/alfresco-content-services/values.yaml
```diff
@@ -307,7 +307,16 @@ alfresco-infrastructure:
     enabled: false
 
 alfresco-search:
-  enabled: true
+  enabled: false
```
andrei-rebegea: true is the default

helm/alfresco-content-services/templates/Rule_04-network-policy-repository.yaml
```diff
@@ -70,11 +72,13 @@ spec:
       - podSelector:
           matchLabels:
              app: {{ template "content-services.shortname" . }}-repository
-      
+
+      {{ if index .Values "alfresco-search" "enabled" }}
```
andrei-rebegea: some of these newly added "if"s start with 
{{- if 

others
{{ if

find out why or be consistent across the file

helm/alfresco-content-services/templates/config-repository.yaml
```diff
@@ -35,9 +35,13 @@ data:
       -Dshare.protocol={{ .Values.externalProtocol | default "http"}}
       -Dshare.host={{ .Values.externalHost | default (printf "%s-share" (include "content-services.shortname" .)) }}
       -Dshare.port={{ .Values.externalPort | default .Values.share.service.externalPort }}
-      -Dalfresco_user_store.adminpassword={{ .Values.repository.adminPassword | default "209c6174da490caeb422f3fa5a7ae634" }}
+      -Dalfresco_user_store.adminpassword={{ .Values.repository.adminPassword | default "209c6174da490caeb422f3fa5a7ae634" }} {{ if index .Values "alfresco-search" "enabled" }}
       -Dsolr.host={{ template "alfresco-search.host" . }}
       -Dsolr.port={{ template "alfresco-search.port" . }}
+      {{ end }} {{ if index .Values "alfresco-search" "external" }}
```
killerboot: If alfresco search is enabled and external is present, then solr.host is set twice, right?

helm/alfresco-content-services/templates/config-repository.yaml
```diff
@@ -35,9 +35,13 @@ data:
       -Dshare.protocol={{ .Values.externalProtocol | default "http"}}
       -Dshare.host={{ .Values.externalHost | default (printf "%s-share" (include "content-services.shortname" .)) }}
       -Dshare.port={{ .Values.externalPort | default .Values.share.service.externalPort }}
-      -Dalfresco_user_store.adminpassword={{ .Values.repository.adminPassword | default "209c6174da490caeb422f3fa5a7ae634" }}
+      -Dalfresco_user_store.adminpassword={{ .Values.repository.adminPassword | default "209c6174da490caeb422f3fa5a7ae634" }} {{ if index .Values "alfresco-search" "enabled" }}
```
killerboot: Can't the "if" clause be on new line?

helm/alfresco-content-services/templates/config-repository.yaml
Epurashu: The external port and host are commented out by default I'm expecting to not be both enable at the same time , but very good point, I will fix this.

Epurashu: Done :)

helm/alfresco-content-services/templates/ingress-repository.yaml
```diff
@@ -10,11 +10,13 @@ metadata:
     nginx.ingress.kubernetes.io/session-cookie-hash: "sha1"
     # Default file limit (1m) check, document(s) above this size will throw 413 (Request Entity Too Large) error
     nginx.ingress.kubernetes.io/proxy-body-size: {{ .Values.repository.ingress.maxUploadSize }}
+    {{- if index .Values "alfresco-search" "enabled" }}
```
mmuller88: If we disable solr we still could use an external one right? So for this case we would need the redirect rules or? It is probably more save to always forbid access to those solr urls with alfresco as a proxy.

PR #174, opened 2019-03-25
[ACA-2293] upgrade ADW to 1.1.0

DenysVuika:
Upgrade to Alfresco Digital Workspace 1.1.0.

- update application image to 1.1.0
- update proxy image to 3.0.0

### DIFF
--- a/helm/alfresco-content-services/values.yaml
+++ b/helm/alfresco-content-services/values.yaml
@@ -331,7 +331,7 @@ alfresco-search:
 alfresco-digital-workspace:
   enabled: true
   ingress:
-    path: /digital-workspace
+    path: /workspace
     applicationBasePath: ./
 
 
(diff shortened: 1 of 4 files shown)


helm/alfresco-content-services/values.yaml
```diff
@@ -331,7 +331,8 @@ alfresco-search:
 alfresco-digital-workspace:
   enabled: true
   ingress:
-    path: /digital-workspace
+    path: /workspace
```
killerboot: Is it required to change any documentation regarding this?

DenysVuika: It was changed in the past for the previous release, we just forgot to update the path for the k8s tests. See https://github.com/Alfresco/acs-deployment/blob/master/docs/docker-compose-deployment.md

PR #116, opened 2018-11-21
Correct port number

anxumalo:

PR #177, opened 2019-03-29
REPO-4300: Update Helm Chart to validate if S3 properties are set and…

cristianturlica:
… not default to empty

   - added checks


killerboot:
This commit fixes #116.

killerboot:
@cristianturlica are the S3 properties only relevant in scope of https://github.com/Alfresco/acs-deployment-aws?

Epurashu:
@killerboot I don't think the S3 properties are relevant only for acs-deployment-aws project, there is an issue for SFS to actually use S3, I think they will make greater use of them in the near future.
Also, you can use alfresco-content-repository-aws image instead of alfresco-content-repository.

killerboot:
@Epurashu I'm ok with the changes. Those seem to be the right thing to have but the Alfresco docker image used here doesn't have the S3 connector, so those are useless by default.
In my opinion SFS needs to be able to use conventional file store in case it is required to deploy this helm chart without using AWS. If the customer actually needs to deploy it in AWS it makes sense to steer them towards https://github.com/Alfresco/acs-deployment-aws.

montgolfiere:
JFYI: ATS-347 will introduce option to helm install ACS deployment with AWS AI (including S3 connector) via alfresco-*-aws images (for repo & share).

PR #178, opened 2019-04-01
REPO-4293: Helm release 1.2.2 for ACS 6.0.1(.2)

ancutam:
Update ACS version to 6.0.1.2 and helm version to 1.2.2.


Epurashu:
If we plan on releasing Chart version 1.2.2 to stable from support/HF/1.0 branch, we should also update helm-chart-releases.md

PR #183, opened 2019-04-10
DEPLOY-655 added hostname option in ingress

mteodori:


killerboot:
@mteodori It seems that ADW can't be accessed any more:
https://bamboo.alfresco.com/bamboo/browse/PLAT-ACSHELM167-2

mteodori:
uhm I see, I should set the hostname there as well, otherwise nginx-ingress ignores the * rules, it's the same problem as in AIS

mteodori:
@killerboot I think I should remove the default to externalHost, there is no way to pass it to a subchart like dw unless we change it to a global value like the ones in the common activiti chart https://github.com/Activiti/activiti-cloud-charts/blob/master/common/values.yaml#L3-L22, plus the issue is also on the ingress of other requirements like alfresco-search and alfresco-insight-zeppelin (is this one still imported? I see it in the README) and I just managed to have the chart of DW changed by @DenysVuika WDYT?

mteodori:
@killerboot removed the default, can we merge until we think about the bigger change (i.e. externalHost to global.externalHost or similar)

PR #188, opened 2019-04-29
multiple instances support

mteodori:
allows having an umbrella chart that installs two instances of ACS with different aliases by using nameOverride on the required components in order to avoid a clash on service DNS names for activemq and search (solr)

### DIFF
--- a/helm/alfresco-content-services/requirements.yaml
+++ b/helm/alfresco-content-services/requirements.yaml
@@ -10,8 +10,8 @@ dependencies:
   repository: https://kubernetes-charts.storage.googleapis.com/
   condition: postgresql.enabled
 - name: alfresco-search
-  version: 0.0.11
-  repository: https://kubernetes-charts.alfresco.com/stable
+  version: 0.0.12
+  repository: https://kubernetes-charts.alfresco.com/incubator
   condition: alfresco-search.enabled
 - name: alfresco-infrastructure
   version: 4.1.0
--- a/helm/alfresco-content-services/values.yaml
+++ b/helm/alfresco-content-services/values.yaml
@@ -336,7 +336,6 @@ alfresco-infrastructure:
     enabled: true
   alfresco-identity-service:
     enabled: false
-
 alfresco-search:
   enabled: true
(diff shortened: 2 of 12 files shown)


helm/alfresco-content-services/requirements.yaml
```diff
@@ -10,8 +10,8 @@ dependencies:
   repository: https://kubernetes-charts.storage.googleapis.com/
   condition: postgresql.enabled
 - name: alfresco-search
-  version: 0.0.11
-  repository: https://kubernetes-charts.alfresco.com/stable
+  version: 0.0.12
+  repository: https://kubernetes-charts.alfresco.com/incubator
```
mmuller88: Not sure if we should pick up from incubator. There can come problem because during the moving from incubator to stable the version is notexisting in incubator anymore because it is a real move and not a copy paste. I would like to insist that we stay on stable

helm/alfresco-content-services/values.yaml
```diff
@@ -433,3 +432,7 @@ s3connector:
 # or uncomment the following line if you don't want/need to pass it as a parameter on every install command :
 # registryPullSecrets: private-repo-registry-secret
 # for more information: https://github.com/Alfresco/alfresco-anaxes-shipyard/blob/master/SECRETS.md
+messageBroker:
```
mmuller88: Let's stick to what we did above and comment those empty value :)

helm/alfresco-content-services/requirements.yaml
mmuller88: That just happened to me in the past that a version wasn't in incubator anymore. As well I think it makes sense to only pick up from stable. For development reason you can override that locally :)

mteodori: I need to ask for a stable release of 0.0.12 as there are some changes there to use nameOverride

helm/alfresco-content-services/values.yaml
mmuller88: As well I guess we need to comment them out

mmuller88:
We are currently working on getting 0.0.12 from search :)

mteodori:
removing the need for two instances in APS so we can close

PR #195, opened 2019-06-12
fix: use templated external* values for repository and share configs

igdianov:
This PR fixes repository and share configs to support templated `externalHost`, `externalPort`, `externalProtocol` custom values required for AAE 7.1 deployment


mteodori:
moved to #197 in same repo in order to merge, this can be closed

PR #197, opened 2019-06-12
fix: use templated external* values for repository and share configs

mteodori:
This PR fixes repository and share configs to support templated externalHost, externalPort, externalProtocol custom values required for AAE 7.1 deployment


mteodori:
extra actions required:
* documentation for templating of external vars
* testing helmAcs.sh script in Alfresco/acs-deployment-aws

mteodori:
added documentation, there is no limitation introduced by this PR on values as the templating need to be activated with "{{ }}" which cannot be used in any of the three variables

PR #210, opened 2019-06-26
Feature/local transform

alandavis:
Changes to support the local transformer changes in repo.

Mainly to bring in the new 'misc' T-Engine and configure up the T-Router so that the repo can access it. The T-Engines are now coming from docker hub rather than quay.io


alandavis:
@killerboot 
Happy to do a release of acs-packaging so I don't need to use latest. What would you suggest as the version (6.2.0-A1)? Green build: https://bamboo.alfresco.com/bamboo/browse/PLAT-ACSHELM180-DI-12/log 

FYI: The number of pods has gone up by 2 again to 24 as we add transform-misc.

docker-compose/docker-compose.yml
```diff
@@ -14,7 +14,7 @@ version: "2"
 
 services:
     alfresco:
-        image: alfresco/alfresco-content-repository:6.1.0
+        image: alfresco/alfresco-content-repository:latest
```
killerboot: I think we need to discuss the possibility of doing EA or RC release with new "transformer" code in.

killerboot:
> Happy to do a release of acs-packaging so I don't need to use latest. What would you suggest as the version (6.2.0-A1)?

Sounds good to me, let's coordinate with the rest of the team in case of anybody else wants to include their changes (unlikely).

docker-compose/docker-compose.yml
montgolfiere: Yes, that would be great - maybe 6.2.0-EA1 to get the ball rolling ?

montgolfiere:
> FYI: The number of pods has gone up by 2 again to 24 as we add transform-misc.

 The idea was to make sure that we deploy by default with at least 2 pods (for reliability as well as testing "cluster" of pods) similar to Repo pods. However, maybe in the future this should be a configurable option to make it easier to scale up or down (to 1 for build/dev envs), eg. based on some minimalistic "profile" ?

montgolfiere:
LGTM - thanks for incorporating first round of T-Router / T-Engine EAs. 

For future, updates, it would be great if we could reduce the number of .env / test.env that need to be updated ?

Issue #208, opened 2019-06-20
yaml indentation bug


torstenwerner:
patch:

    Index: helm/alfresco-content-services/templates/deployment-transform-router.yaml
    IDEA additional info:
    Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
    <+>UTF-8
    ===================================================================
    --- helm/alfresco-content-services/templates/deployment-transform-router.yaml	(revision c6781aa1e0e6bf5b5c871e028cab1f4a490ebb5d)
    +++ helm/alfresco-content-services/templates/deployment-transform-router.yaml	(revision 309128ffecc808f94b86d7f1cfc8d0e36c2b7d5e)
    @@ -57,17 +57,17 @@
                 httpGet:
                   path: /actuator/health
                   port: {{ .Values.transformrouter.image.internalPort }}
    -              initialDelaySeconds: {{ .Values.transformrouter.livenessProbe.initialDelaySeconds }}
    -              periodSeconds: {{ .Values.transformrouter.livenessProbe.periodSeconds }}
    -              failureThreshold: 1
    -              timeoutSeconds: {{ .Values.transformrouter.livenessProbe.timeoutSeconds }}
    +            initialDelaySeconds: {{ .Values.transformrouter.livenessProbe.initialDelaySeconds }}
    +            periodSeconds: {{ .Values.transformrouter.livenessProbe.periodSeconds }}
    +            failureThreshold: 1
    +            timeoutSeconds: {{ .Values.transformrouter.livenessProbe.timeoutSeconds }}
               readinessProbe:
                 httpGet:
                   path: /actuator/health
                   port: {{ .Values.transformrouter.image.internalPort }}
    -              initialDelaySeconds: {{ .Values.transformrouter.readinessProbe.initialDelaySeconds }}
    -              periodSeconds: {{ .Values.transformrouter.readinessProbe.periodSeconds }}
    -              timeoutSeconds: {{ .Values.transformrouter.readinessProbe.timeoutSeconds }}
    +            initialDelaySeconds: {{ .Values.transformrouter.readinessProbe.initialDelaySeconds }}
    +            periodSeconds: {{ .Values.transformrouter.readinessProbe.periodSeconds }}
    +            timeoutSeconds: {{ .Values.transformrouter.readinessProbe.timeoutSeconds }}
               {{- if .Values.ai.enabled }}
               volumeMounts:
               - name: config-volume

killerboot:
[...]

Issue #207, opened 2019-06-19
broken configuration of transform-router


torstenwerner:
The imagePullSecrets do not fall back to the secrets referenced in the serviceaccount by default. Here is the patch:

    Index: helm/alfresco-content-services/templates/deployment-transform-router.yaml
    IDEA additional info:
    Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
    <+>UTF-8
    ===================================================================
    --- helm/alfresco-content-services/templates/deployment-transform-router.yaml	(date 1560973799000)
    +++ helm/alfresco-content-services/templates/deployment-transform-router.yaml	(date 1560974160000)
    @@ -17,9 +17,11 @@
             release: {{ .Release.Name }}
             component: transformrouter
         spec:
    +      {{- if .Values.registryPullSecrets }}
           # only set this secret if a private docker registry variable is defined
           imagePullSecrets:
             - name: {{ .Values.registryPullSecrets }}
    +      {{- end }}
           affinity:
             podAntiAffinity:
               preferredDuringSchedulingIgnoredDuringExecution:
    @@ -76,4 +78,4 @@
           - name: config-volume
             configMap:
               name: {{ template "alfresco.shortname" . }}-transform-routes-configmap
    -      {{- end }}
    \ No newline at end of file
    +      {{- end }}

killerboot:
Thanks for raising this @torstenwerner

@montgolfiere Please take a look at this issue.

danielfz:
I getting an error just doing "docker-compose up". Not sure if it's related to this.

`Pulling transform-router (quay.io/alfresco/alfresco-transform-router:1.1.0)...
ERROR: unauthorized: access to the requested resource is not authorized`

ict-singapore:
> I getting an error just doing "docker-compose up". Not sure if it's related to this.
> 
> `Pulling transform-router (quay.io/alfresco/alfresco-transform-router:1.1.0)... ERROR: unauthorized: access to the requested resource is not authorized`

Have you received the **quay.io** credentials from Alfersco support. Only then you can download this image to build the transform-router container. More details are [here](https://docs.alfresco.com/6.1/tasks/deploy-docker-compose.html)

montgolfiere:
Closing as per comment above.

See also https://github.com/Alfresco/acs-deployment/blob/master/README.md#other-information

PR #212, opened 2019-07-03
fixed validation error on resources with helm-2.14/kubectl-1.14

mteodori:
launch with helm 2.14 which uses the kubectl 1.14 lib:

Error: error validating "": error validating data: [ValidationError(Deployment.spec.template.spec.containers[0].livenessProbe): unknown field "limits" in io.k8s.api.core.v1.Probe, ValidationError(Deployment.spec.template.spec.containers[0].livenessProbe): unknown field "requests" in io.k8s.api.core.v1.Probe]

### DIFF
--- a/helm/alfresco-content-services/templates/deployment-transform-router.yaml
+++ b/helm/alfresco-content-services/templates/deployment-transform-router.yaml
@@ -1,5 +1,5 @@
 apiVersion: extensions/v1beta1
-kind: Deployment
+apiVersion: apps/v1
 metadata:
   name: {{ template "alfresco.shortname" . }}-router
   labels:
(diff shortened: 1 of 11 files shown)


mteodori:
fixes also #207 and #208

killerboot:
The issue will be fixed and tracked in https://issues.alfresco.com/jira/browse/REPO-4513

helm/alfresco-content-services/templates/deployment-transform-router.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: extensions/v1beta1
-kind: Deployment
+apiVersion: apps/v1
```
gavincornwell, line 2: Is this a typo? I presume we need to keep the "kind: Deployment" and change the previous line!?

mteodori: yes definitely

PR #213, opened 2019-07-11
REPO-4523 : Use dynamic provisioning to create Persistent Volumes

constantinpopa10:
- modified helm-deployment-aws_kops by:
      -> adding a table of contents
      -> adding instructions to create a NFS client provisioner
      -> updating the ACS install steps to use the latest alfresco-infrastructure chart that supports dynamic provisioning

### DIFF
--- a/docs/helm-deployment-aws_kops.md
+++ b/docs/helm-deployment-aws_kops.md
@@ -1,3 +1,36 @@
+Table of Contents
+=================
+
+   * [Alfresco Content Services Deployment with Helm on AWS using Kops](#alfresco-content-services-deployment-with-helm-on-aws-using-kops)
+      * [Prerequisites](#prerequisites)
+         * [Prepare and configure Kops](#prepare-and-configure-kops)
+         * [Setting up Kubernetes cluster on AWS with Kops](#setting-up-kubernetes-cluster-on-aws-with-kops)
+         * [Validate cluster](#validate-cluster)
+         * [Upgrading a cluster (optional)](#upgrading-a-cluster-optional)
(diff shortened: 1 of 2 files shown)


killerboot:
@constantinpopa10 If alfresco-infrastructure:5.1.1 is included in the ACS chart, will it still be required to install infrastructure separately?

constantinpopa10:
@killerboot  Yes, if ACS were to use 5.1.1, you wouldn't need to install the persistence on its own.

Only the persistence part of the alfresco-infrastructure:5.1.1 is installed separately.
ACS still uses the infrastructure:4.1.0 to bring activemq for example.

killerboot:
@constantinpopa10 I think it is better to update the infrastructure first. A stable version should exist before doing that.

constantinpopa10:
@killerboot if I understood correctly from you, we'll keep this PR and this [one](https://git.alfresco.com/Repository/acs-k8s-cluster/merge_requests/16)   opened until the **alfresco-infrastructure** releases a stable 5.1.1 version.

This means the acs-deployment [build](https://bamboo.alfresco.com/bamboo/browse/PLAT-ACSHELM) will remain unstable until then. Nevertheless as long as the acs-deployment build plan is using the [feature branch with the fix](https://git.alfresco.com/Repository/acs-k8s-cluster/tree/feature/REPO-4523_dynamic_volumes) it should be ok.

killerboot:
@constantinpopa10 agreed. I will raise the update of the chart in a separate story as a prerequisite to this work.

constantinpopa10:
I've updated the PR to use the alfresco-infrastructure:4.1.1 which adds dynamic provisioning.

ancutam:
Look very good!

PR #233, opened 2019-09-06
update sync service to final release 3.1.2

ciprianchelaru:


ancutam:
@ciprianchelaru There is still a reference to the RC versions of the sync-services chart [helm/alfresco-content-services/requirements.yaml](https://github.com/Alfresco/acs-deployment/blob/support/SP/2.N/helm/alfresco-content-services/requirements.yaml#L25)

PR #237, opened 2019-09-13
Changes to pick up ACS 6.2.0-A7

alandavis:
Includes gytheio 0.10.1


ancutam:
Please update here as well https://github.com/Alfresco/acs-deployment/blob/feature/repo-4639_tika/docker-compose/docker-compose.yml#L15

PR #249, opened 2019-09-26
REPO-4670: Update helm chart docker image information

amohammedalfresco:
- Extend [docs/helm-chart-releases.md](https://github.com/Alfresco/acs-deployment/blob/master/docs/helm-chart-releases.md) file with docker image information which is part of ACS Helm chart.

### DIFF
--- a/docs/helm-chart-releases.md
+++ b/docs/helm-chart-releases.md
@@ -1,22 +1,22 @@
 ## Stable Helm Chart Releases
 
-|Chart version|ACS version|
-|:---:|:---:|
-|1.0.0|6.0.0|
-|1.0.1|6.0.0|
-|1.0.2|6.0.0|
-|1.0.3|6.0.0|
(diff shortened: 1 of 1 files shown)


amohammedalfresco:
Thanks @killerboot! The (MNT-20924)[https://issues.alfresco.com/jira/browse/MNT-20924] is requesting for this information against each ACS helm chart release.

montgolfiere:
Also, means another place to remember (or forget) to maintain docker image versions (in addition to helm chart values & docker-compose files) ?

ancutam:
That's a lot of work, Abdul! But unless it is generated somehow, I don't see how we will maintain it.

amohammedalfresco:
It seems maintaining this piece of information is going to be a task of its own due to a need of any automation and also duplication of information.

Hence, for the time being I will close this PR and update the requester about it.

PR #279, opened 2019-10-31
fix(MNT-21042): added rollingUpdate values with sensible defaults

igdianov:
This PR fixes https://issues.alfresco.com/jira/browse/MNT-21042

- [x] Added rolling update strategy deployment configuration `maxUnavailble=0` to ensure no unvailable pod replicas with `maxSurge=1`

- [x] Added Container `PreStop` lifecycle hook to wait in order to properly drain existing connections before pod shutdown
    
repository-deployment.yaml  
```
spec:   
  replicas: {{ .Values.repository.replicaCount }}  
  # Sensible default values for rolling updates
  strategy: 
    type: RollingUpdate 
    rollingUpdate:  
      maxSurge: {{ .Values.rollingUpdateMaxSurge | default 1 }} 
      maxUnavailable: {{ .Values.rollingUpdateMaxUnavailable | default 0 }}     
  template: 
    metadata:   
      containers:   
        - name: {{ .Chart.Name }}   
          # sleep to drain any existing connections
          lifecycle:    
            preStop:    
              exec: 
                command: ["/bin/bash", "-c", "sleep 20"]    
```

Resolution: Rolling update starts  new pod and initiates old pod shutdown only after existing connections are drained to achieve zero donwtime update

### DIFF
--- a/helm/alfresco-content-services/templates/deployment-ai-transformer.yaml
+++ b/helm/alfresco-content-services/templates/deployment-ai-transformer.yaml
@@ -13,6 +13,11 @@ metadata:
     component: transformers
 spec:
   replicas: {{ .Values.aiTransformer.replicaCount }}
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxSurge: {{ .Values.rollingUpdateMaxSurge | default 1 }}
+      maxUnavailable: {{ .Values.rollingUpdateMaxUnavailable | default 0 }}    
   template:
     metadata:
       labels:
--- a/helm/alfresco-content-services/templates/deployment-filestore.yaml
+++ b/helm/alfresco-content-services/templates/deployment-filestore.yaml
@@ -12,6 +12,11 @@ metadata:
     component: transformers
 spec:
   replicas: {{ .Values.filestore.replicaCount }}
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxSurge: {{ .Values.rollingUpdateMaxSurge | default 1 }}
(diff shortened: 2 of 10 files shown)


killerboot:
I don't think that rolling update of content-repository is supported if the DB schema changed between versions or patching of data in the tables is required.
cc @skopf

skopf:
@killerboot has a valid point about rolling updates not being available for all ACS updates.
But I assume that for most of the upgardes performed with this mechanism, customers are building new custom images of the same base ACS version and are using this to deploy a new version of their AMP instead of a new ACS version.
And also with the recent versions, our DB schema has been pretty stable and we have been able to allow rolling updates.

Also looking at this page
https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#strategy
the "RollingUpdates" startegy is the default anyway. This PR just fixes the parameters of the RollingUpdate.

@amohammedalfresco With these changes in place, is it still possible to enforce a Recreate update for specific deployments, especially the repository deployment?
I think we should introduce something like
```
spec:
  replicas: {{ .Values.repository.replicaCount }}
  strategy:
    type: {{ .Values.repository.updateStrategyType | default "RollingUpdate" }}
```
(not sure if the above is valid)
For now, we only need this for the repository. This is the only deployment in our chart that has a database and might require a Recreate update.

amohammedalfresco:
I have a small issue committing in the fork branch `igdianov:fix-MNT-21042-rolling-upgate` directly for @skopf's recommendation.  Hence, I will merge the PR into an alfresco equivalent branch (`Alfresco:fix-MNT-21042-rolling-update`) and then commit the recommendation.

PR #278, opened 2019-10-31
fix(MNT-21041): added checksum annotations for configmaps

igdianov:
This PR fixes https://issues.alfresco.com/jira/browse/MNT-21041

Added checksum annotation to deployment templates in order to trigger pod restart on any changes in config map values when running `helm upgrade` command, i.e.	
	
deployment-share.yaml:
```
spec:	
  template:	
    metadata:	
      annotations:	
        checksum/config: {{ include (print $.Template.BasePath "/config-share.yaml") . | sha256sum }}	
```
	
deployment-repository.yaml	
```
spec:	
  template:	
    metadata:	
      annotations:	
        checksum/config: {{ include (print $.Template.BasePath "/config-repository.yaml") . | sha256sum }}	
```	
	
Resolition: Helm upgrade will trigger pod restart when any configurations values in config map  have changed.

PR #296, opened 2020-01-02
REPO-4916 MNT-21041: ACS Helm upgrade to update configuration values does not restart affected pods

amohammedalfresco:
- #278 
- Enhance the PR by covering secrets (i.e. if passwords or aws access key/secrets are changed then it should trigger a new repository pod with updated credentials).


amohammedalfresco:
> Is it required to make any changes to the dependencies, for example search?

I believe yes.  In general, we should make changes to dependent (alfresco-owned) charts.

killerboot:
@sridharvellingiri Please take a look at this PR. Similar changes may be required in Search and Discovery charts.

amohammedalfresco:
Latest bamboo build with updated Search/Zepplin helm charts containing supporting changes.

PR #297, opened 2020-01-10
MNT-21042: ACS Helm upgrade fails to perform rolling update with zero downtime

amohammedalfresco:
- #279 
- Parameterise `strategy` annotation's `type` option for Repository deployment as `Recreate`
- `rollingUpdate` are defined as global parameters to enable Alfresco dependent helm charts can inherit values when ACS helm chart is updated.

### DIFF
--- a/helm/alfresco-content-services/templates/deployment-repository.yaml
+++ b/helm/alfresco-content-services/templates/deployment-repository.yaml
@@ -12,10 +12,11 @@ metadata:
 spec:
   replicas: {{ .Values.repository.replicaCount }}
   strategy:
-    type: {{ .Values.repository.updateStrategyType | default "RollingUpdate" }}
-    rollingUpdate:
-      maxSurge: {{ .Values.rollingUpdateMaxSurge | default 1 }}
-      maxUnavailable: {{ .Values.rollingUpdateMaxUnavailable | default 0 }}    
+    type: Recreate
+    rollingUpdate: null
+    # rollingUpdate:
+    #   maxSurge: {{ .Values.rollingUpdateMaxSurge | default 1 }}
+    #   maxUnavailable: {{ .Values.rollingUpdateMaxUnavailable | default 0 }}    
   template:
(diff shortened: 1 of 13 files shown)


helm/alfresco-content-services/values.yaml
```diff
@@ -11,6 +11,8 @@
 
 repository:
   replicaCount: 2
+  # MNT-21042: enforce a Recreate update for the repository deployment
```
killerboot: Not sure if you need a reference to MNT here

amohammedalfresco: Thanks @killerboot, I just removed that reference.

helm/alfresco-content-services/templates/deployment-repository.yaml
```diff
@@ -11,6 +11,12 @@ metadata:
     component: repository
 spec:
   replicas: {{ .Values.repository.replicaCount }}
+  strategy:
+    type: Recreate
+    rollingUpdate: null
```
amohammedalfresco: @skopf: I believe as per [#24198](https://github.com/kubernetes/kubernetes/issues/24198#issuecomment-212730719), this would be an example of how we can implement `Recreate` strategy type.

skopf: Ah, I see.
So when the live system has ```strategy.type==RollingUpdate```, then we cannot simply set it to ```"Recreate"```. We additionally need to ```null``` the ```strategy.rollingUpdate`` structure.
But with this change, we now always use ```strategy.type==Recreate``` without the possibility to overwrite it.
How about:
```
  strategy:
    type: {{ .Values.repository.updateStrategyType | default "RollingUpdate" }}
    rollingUpdate:  {{ .Values.repository.updateStrategyRollingUpdate }}
```
Would this be possible? Can we bring a whole dictionary from the values? Or is it only possible for scalar values?

PR #306, opened 2020-03-04
ATS-651 : AIS: Release Intelligence Services 1.1.2

tzclucian:
- upgradin AIS


tzclucian:
@montgolfiere I don't think there's anything to upgrade [here](https://github.com/Alfresco/acs-deployment/blob/master/helm/alfresco-content-services/ai-reference-values.yaml). Am I correct?

PR #308, opened 2020-03-27
MNT-20671: Alfresco DBP using the AWS EKS deployment method, inbound email port is unavailable for ACS

amohammedalfresco:
- Configure email inbound/outbound for ACS in helm chart via `nginx-ingress`


helm/alfresco-content-services/templates/config-repository.yaml
```diff
@@ -61,6 +61,45 @@ data:
...
+      -Dmail.smtps.starttls.enable={{ .Values.mail.smtps.starttls.enable }}
+      -Dimap.server.enabled={{ .Values.imap.server.enabled }}
+      -Dimap.server.port={{ .Values.imap.server.port }}
+      -Dimap.server.host={{ .Values.imap.server.host }}
+      -Dimap.server.imap.enabled={{ .Values.imap.server.imap.enabled }}
+      -Dimap.server.imaps.enabled={{ .Values.imap.server.imaps.enabled }}
+      -Dimap.server.imaps.port={{ .Values.imap.server.imaps.port }}
+      -Dimap.mail.from.default={{ .Values.imap.mail.from.default }}
+      -Dimap.mail.to.default={{ .Values.imap.mail.to.default }}
+      -Dcsrf.filter.enabled=false
```
gavincornwell, line 100: Do you intend to put the "csrf.filter.enabled" and "system.usages.enabled" flags inside the  "{{- if" statement?

amohammedalfresco: I've seen these 2 params in [prerequisite-for-acs-60-using-docker-images](https://github.com/Alfresco/alfresco-tas-email#prerequisite-for-acs-60-using-docker-images) and rendering it inside the if/end block [here](https://github.com/Alfresco/acs-deployment/blob/8fa2f04fdf440c3efdee069e1458fff50fcc344b/helm/alfresco-content-services/templates/config-repository.yaml#L64-L102)

gavincornwell: Hi Abdul, I reached out to Alex to double check on this, his response was: 

> Hi, disabling the csrf filter is a security problem.
> it should be configured with valid origin and referer instead

So as I suspected we're going to need to address this as we can't introduce a security hole when someone wants to use email.

PR #310, opened 2020-04-16
REPO-5079, ATS-707: Use all-in-one T-Engine and fix tests

montgolfiere:
- update docker-compose references to use single Core AIO T-Engine (rather than 5x T-Engines)
-- bump to T-Engines 2.2.0-A4
-- also initial update to T-Router config
-- update transform-core-aio service name (to be consistent with other packaging/deployment PRs)
-- increase Xmx for single
-- update docker-compose puml diagram

- update helm chart
-- bump values to T-Engines 2.2.0-A4
-- note: leave ACS Enterprise Deployment helm charts using 5x Core T-Engines (independently scalable)


alandavis:
WIP: Going to switch back to a separate .url for each of the legacy T-Engines so the chart in acs-deployment can point legacy transforms to separate T-Engines. Otherwise we would need a AIO there in addition for testing.

montgolfiere:
Ready for review (to merge to master).

PR #313, opened 2020-04-24
ATS-739: Update ACS Deployment to ATS 1.2.0-A2

montgolfiere:
- T-Router 1.2.0-A2 with T-Engines 2.2.0-A5
- change to single CORE_AIO_URL (ATS-736)

### DIFF
--- a/docker-compose/docker-compose.yml
+++ b/docker-compose/docker-compose.yml
@@ -52,25 +52,12 @@ services:
 
     transform-router:
         mem_limit: 512m
-        image: quay.io/alfresco/alfresco-transform-router:1.1.1
+        image: quay.io/alfresco/alfresco-transform-router:1.2.0-A2
         environment:
           JAVA_OPTS: " -Xms256m -Xmx512m"
           ACTIVEMQ_URL: "nio://activemq:61616"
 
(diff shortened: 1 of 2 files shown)


docker-compose/docker-compose.yml
```diff
@@ -52,25 +52,12 @@ services:
...
-
-          LIBREOFFICE_URL : "http://transform-core-aio:8090"
-          LIBREOFFICE_QUEUE : "org.alfresco.transform.engine.aio.acs"
-
-          TIKA_URL : "http://transform-core-aio:8090"
-          TIKA_QUEUE : "org.alfresco.transform.engine.aio.acs"
-
-          MISC_URL : "http://transform-core-aio:8090"
-          MISC_QUEUE : "org.alfresco.transform.engine.aio.acs"
+          CORE_AIO_URL : "http://transform-core-aio:8090"
```
PunchedChimera, line 61: Should the CORE_AIO_QUEUE variable be included here?

montgolfiere: No, it's not needed.

As with previous releases (when using individual x5 T-Engines) it's already pre-configured with a default to match Core AIO T-Engine default :-)

PR #344, opened 2020-06-08
ACS-270 Removing and Adding SyncService chart to acs-deployment subchart

sridharvellingiri:


sridharvellingiri:
Latest bamboo job for the change --> https://bamboo.alfresco.com/bamboo/browse/PLAT-ACSHELM322-DCT-10

helm/alfresco-content-services/values.yaml
```diff
@@ -598,6 +598,9 @@ global:
     - quay-registry-secret
 
 alfresco-sync-service:
-  enabled : true
+  syncservice:
```
gavincornwell, line 601: Do we need to add an extra level here?

helm/alfresco-content-services/charts/alfresco-sync-service/charts/postgresql/Chart.yaml
```diff
@@ -0,0 +1,17 @@
+description: Object-relational database management system (ORDBMS) with an emphasis
```
gavincornwell, line 1: I'm not sure I understand why we have to bring the whole postrges chart into ACS, surely we don't want to maintain the postgres chart!?

Can we not define postgres as a requirement of the ACS chart?

helm/alfresco-content-services/values.yaml
sridharvellingiri: Yes, by default its true and just added extra level for optional to disable the service which was already there.

helm/alfresco-content-services/charts/alfresco-sync-service/charts/postgresql/Chart.yaml
sridharvellingiri: its in requirement actually but during deployment, the sync postgres is not getting deployed because it requires "helm dep up" inside SyncService subchart to download the chart.
**Note:** we also need to update acs-k8s deployment script to run dependencies for subchart.
May be we can remove postgres subchart and add the steps in README???

mteodori: yep definitely we do not want the postgres chart code we don't own in as @gavincornwell suggests we can have a second postgresql in the acs chart requirements with an alias like postgresql-sync-service

sridharvellingiri: Postgresql chart has been defined one level up in acs requirements. Here is the latest bamboo job for the change --> https://bamboo.alfresco.com/bamboo/browse/PLAT-ACSHELM322-12

PR #397, opened 2020-07-31
Feature/acs-362 dsync helm chart test

sridharvellingiri:


.travis.yml
```diff
@@ -64,8 +64,17 @@ after_failure: |
...
+stages:
+  - test
+
 jobs:
   include:
+    - name: Lint for subcharts
+      stage: test
+      script: |
+        helm lint helm/alfresco-content-services/charts/activemq
+        helm lint helm/alfresco-content-services/charts/alfresco-search
```
gavincornwell, line 76: There's also a sub-sub-chart for the insight engine should we lint that as well?

sridharvellingiri: Insight-engine is part of search helm chart, you mean insight-zeppelin helm chart?

sridharvellingiri: added lint for insight-zeppelin

PR #415, opened 2020-09-10
support/SP/4.N: Upgrade Sync Service to 3.3.3.1 after hotfix release

SimonaCarp:


gavincornwell:
Hi Simona, the old support/SP/4.N branches are not being maintained anymore and the Bamboo builds have been disabled so the chart will never get published.

We now have a set of values files on the master branch, one of them is for 6.2.x. Can you please make the change there and close this pull request. 

Thanks,

Gavin

PR #419, opened 2020-09-14
OPSEXP-105: Fix and stabilise build

gavincornwell:


.travis.yml
```diff
@@ -238,11 +238,25 @@ _helm_deploy: &helm_deploy
...
+      ACS tests failed, exiting
+      exit 1
+    fi
     # run sync service checks
     if [[ "$values_file" != "helm/alfresco-content-services/6.0.N_values.yaml" ]] && [[ "$values_file" != "helm/alfresco-content-services/community_values.yaml" ]]; then
       docker run -a STDOUT --volume $PWD/test/postman/helm:/etc/newman --network host postman/newman_alpine33:3.9.2 run "sync-service-test-helm-collection.json" --global-var "protocol=https" --global-var "url=$namespace.dev.alfresco.me"
+      SYNC_TEST_RESULT=$?
+      if [[ "${SYNC_TEST_RESULT}" == "1" ]]; then
+        Sync tests failed, exiting
+        exit 1
```
sergiuv2020, line 258: Dont exit with 1 here @gavincornwell ill will kill the session again, how will the cleanup run then?

sergiuv2020 (changes requested):

.travis.yml
sergiuv2020: https://docs.travis-ci.com/user/job-lifecycle you need to use -> TRAVIS_TEST_RESULT

gavincornwell: I did read about exit 1 not being a good idea after implementing this but we use it in multiple places in the script already :-( 

Also, I tried to use TRAVIS_TEST_RESULT and it doesn't work as advertised, it was always empty, never 0 or 1.  However, even if did work what do we do with the result if we can't exit!?

I'm beginning to hate Travis ;-)

lsuciu (approved):

PR #427, opened 2020-09-29
OPSEXP-555: keystore additions

sergiuv2020:
Added new keystore props to the helm deployments, also altered the values files for older version but kept the metadata pass and key pass configurable and secured with secrets

### DIFF
--- a/docker-compose/docker-compose.yml
+++ b/docker-compose/docker-compose.yml
@@ -12,15 +12,19 @@ version: "2"
 
 services:
     alfresco:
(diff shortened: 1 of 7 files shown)


gavincornwell (approved):
LGTM

gavincornwell:
Thinking about it, do we need to make the same changes to the older docker compose files too?

sergiuv2020:
> Thinking about it, do we need to make the same changes to the older docker compose files too?

Not yet as it seems for now. Maybe after a subsequent release of those.

PR #432, opened 2020-10-08
OPSEXP-574: Update search dependencies

gavincornwell:


helm/alfresco-content-services/charts/alfresco-sync-service/Chart.yaml
```diff
@@ -1,5 +1,4 @@
 apiVersion: v1
-appVersion: 3.3.3.1
```
killerboot, line 2: Is this related to search?

helm/alfresco-content-services/charts/alfresco-sync-service/Chart.yaml
```diff
@@ -9,4 +8,4 @@ keywords:
 maintainers:
 - name: Alfresco Apps Repo Team
 name: alfresco-sync-service
-version: 3.0.6
+version: 3.0.7
```
killerboot, line 11: Is this related to search?

killerboot (commented):
Is it also worth removing the removed property "solr.content.dir" from here: https://github.com/Alfresco/acs-deployment/blob/0ffb1bece79776e93f66b4b52956305c24e33e25/helm/alfresco-content-services/charts/alfresco-search/values.yaml#L42
This property was removed in 2.0, see https://docs.alfresco.com/search-enterprise/concepts/external-properties-solr.html. But it is still present in older versions.

helm/alfresco-content-services/charts/alfresco-sync-service/Chart.yaml
gavincornwell: Fair point, no it's not, I will revert from this PR.

helm/alfresco-content-services/charts/alfresco-sync-service/Chart.yaml
gavincornwell: Fair point, no it's not, I will revert from this PR.

gavincornwell:
Thank you also for highlighting the "solr.content.dir" property, we will investigate how to conditionally include that for older versions.

killerboot (approved):

PR #367, opened 2020-06-26
OPSEXP-115 & ACS-75: Update charts to support 1.16 and helm3 and move everything to travis

sergiuv2020:
https://issues.alfresco.com/jira/browse/OPSEXP-75 - Move ACS deployment build to Travis
https://issues.alfresco.com/jira/browse/OPSEXP-115 - Support helm 3 and k8s 1.16+ for ACS 7

Added activemq chart along with sync and search
Changed to latest postgres chart with 11.7 version.
Adapted values for sync.
Fixed persistence test collection.
Created travis build for :
- docker-compose
- master & SP2.N on helm3 and k8s 1.16

Chart Version and Travis Branch condition will be changed after the review is done.

PR #437, opened 2020-10-15
OPSEXP-586 Validate helm charts per k8s version

mteodori:
see OPSEXP-586


mteodori:
@sergiuv2020 still not ok for 1.17 with the changes in #367

helm/alfresco-content-services/README.md
```diff
@@ -1,3 +1,354 @@
-# Alfresco Content Services (ACS) Helm Chart
+# alfresco-content-services
```
gavincornwell, line 1: Whilst I appreciate generating docs is a good thing to do this repo is used for end-user documentation at the moment and this overwrites one of the pages we need to guide users to the correct place. Furthermore, this information is already included in a different location.
We are starting a project with the docs team to address this but in the meantime I need to keep the docs as they are and address the auto-generated docs in https://issues.alfresco.com/jira/browse/OPSEXP-435

helm/alfresco-content-services/templates/config-share.yaml
```diff
@@ -10,7 +10,6 @@ data:
   {{ $alfprotocol := tpl (.Values.externalProtocol | default "http") $ }}
   {{ $alfport := tpl (.Values.externalPort | default .Values.repository.service.externalPort | toString) $ }}
   {{ $alfproxy := printf "-Dalfresco.proxy=%s://%s:%s" $alfprotocol $alfhost $alfport }}
-  CATALINA_OPTS: {{ index .Values "share" "environment" "CATALINA_OPTS" }}
```
gavincornwell, line 13: This doesn't look correct, there are valid JVM options specified in the values file so not sure why this has been removed.

helm/alfresco-content-services/templates/deployment-repository.yaml
```diff
@@ -13,7 +13,6 @@ spec:
   strategy:
     {{- if eq .Values.repository.strategy.type "Recreate" }}
     type: {{ .Values.repository.strategy.type }}
-    rollingUpdate: null
```
gavincornwell, line 16: This was added for a reason to address a k8s issues (that granted may have been fixed by now). See https://issues.alfresco.com/jira/browse/MNT-21042 and https://github.com/Alfresco/acs-deployment/commit/09b41c0c2eddad5985238ceb931d564d881ff1c5 for more details.

helm/alfresco-content-services/values.yaml
```diff
@@ -469,6 +468,8 @@ postgresql:
   enabled: true
   nameOverride: postgresql-acs
   imageTag: "11.7.0"
+  commonAnnotations:
```
gavincornwell, line 471: Do you know why we need to add this?

gavincornwell (changes requested):
There are a few things to check/change in here before merging and the helm-docs hook is overwriting a README page we need.

helm/alfresco-content-services/README.md
mteodori: that README is basically empty ATM though and if you move that info in the description field of Chart.yaml you have it there so I think we could go for it

helm-docs is less smart than terraform-docs as it overwrites with a template rather than using a placeholder but maybe it's in the docs and I missed that

helm/alfresco-content-services/templates/config-share.yaml
mteodori: if you generate the template it's duplicated as the environment dict is dumped after, I validated the output with the k8s schema

helm/alfresco-content-services/templates/deployment-repository.yaml
mteodori: it's an invalid value I think the issue got it wrong, rollingUpdate section can be present only when type is RollingUpdate not Recreate

helm/alfresco-content-services/values.yaml
mteodori: it's an harmless fake placeholder, there is an error in the postgresql chart where it still prints the annotations: empty value in ingress and services which does not validate, so in case you have a kubernetes admission controller active which checks for validation the chart does not deploy, with this change the output file validates fully

helm/alfresco-content-services/README.md
gavincornwell: Yes, we have the table of configuration options on this page: https://github.com/Alfresco/acs-deployment/blob/master/docs/helm/README.md so I don't really want it duplicated. 
Can we disable the docs generation for now and do it in OPSEXP-435?

helm/alfresco-content-services/templates/config-share.yaml
gavincornwell: Ah, gotcha, nice catch in the case!

helm/alfresco-content-services/templates/deployment-repository.yaml
gavincornwell: Good point!

helm/alfresco-content-services/values.yaml
gavincornwell: OK, thanks.

gavincornwell (approved):
Looks good to me now and we have a green build with all tests run.

PR #446, opened 2020-10-20
OPSEXP-586 use shared travis config

mteodori:
replaced custom tool install lines with shared configs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-slanguage: python
+language: python
 python:
   - "3.8.2"
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,8 +20,8 @@ stages:
 
 import:
   - source: Alfresco/alfresco-process-tools:.travis.kubernetes_install.yml@master
-  - source: Alfresco/alfresco-process-tools:.travis.helm.yml@master
-  - Alfresco/alfresco-process-tools:.travis.kubepug_install.yml@master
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,4 +1,4 @@
-language: python
+slanguage: python
```
gavincornwell, line 1: Typo?

.travis.yml
```diff
@@ -20,6 +20,8 @@ stages:
 
 import:
   - source: Alfresco/alfresco-process-tools:.travis.kubernetes_install.yml@master
+  - source: Alfresco/alfresco-process-tools:.travis.helm.yml@master
```
gavincornwell, line 23: Doing this breaks our build, see https://travis-ci.com/github/Alfresco/acs-deployment/builds/191249066 so it was omitted previously purposely. 

The publish task runs too early. I raised a support ticket with Travis previously, there is a way to fix the order but we have also customised the publish process slightly so can't use the generic one.

If we can decide on a common way of publishing charts we might be able to fix it.

gavincornwell (changes requested):
The kubepug related changes look good but importing the common Helm script breaks the build :-(

.travis.yml
mteodori: definitely thanks

.travis.yml
mteodori: got it wrong sorry, it should have been just install

gavincornwell (approved):
LGTM

PR #447, opened 2020-10-22
OPSEXP-592: Fix SOLR issues

gavincornwell:
- Changed default memory limits for search to 2G
- Also changed the URL to use for the liveness probe
- Fixed startup issue in 6.0.N docker compose
- Updated DfD guide to reduce the search memory


nilsSure (approved):
LGTM

PR #449, opened 2020-11-02
OPSEXP-599: db max_connections not taken into consideration

sergiuv2020:
Bitnami helm chart is expecting different values for extra posgresql configuration ->
https://github.com/bitnami/charts/blob/master/bitnami/postgresql/values.yaml#L201


gavincornwell (approved):
LGTM

PR #450, opened 2020-11-02
OPSEXP-559: Add percentage ram settings to java apps

sergiuv2020:
Changed hard coded memory settings to use Min/MaxRAMPercentage. Used 50%/80% to give the OS some memory.

More research is required before adding these settings to the search containers, also made sure all search containers don't set any memory so it can be handled by the SOLR scripts.


nilsSure (approved):
LGTM

PR #452, opened 2020-11-09
MNT-21967 MNT-21800 - Update ACS to v6.2.2.5

evasques:


jottley (approved):

gavincornwell:
@evasques @jottley Just to let you know that the build failed after merging to master. It looks like the Docker image is not available yet!? https://travis-ci.com/github/Alfresco/acs-deployment/jobs/431233928

FYI, if you create a branch prefixed with "feature/" you'll get a Travis build for the branch that you can use to check before merging.

evasques:
@gavincornwell can't see how, I only updated the helm charts after i got the all-greens on acs-packaging release. I can see the image in quay from 3h ago: https://quay.io/repository/alfresco/alfresco-content-repository?tab=tags
Can it be something else that I am missing? This is the first release I am making from the new structure...

gavincornwell:
@evasques @jottley The problem is that the deployment artefacts are expecting to find the alfresco repository image in DockerHub, not in quay.io. 

This might be fallout from the latest acs-packaging changes following the consolidation of the GitHub repositories. I'm not aware of any decisions to not push the images to DockerHub so I think they should be there.

Can you please revert this change until we have a working solution, preferably on a branch first.

Also, I see you use a "fix/*" pattern for your branches. Feel free to update the Travis build script to include this pattern if the better matches your processes.

alandavis:
@gavincornwell a decision with Mario and others was made not to push enterprise images to dockerhub. This is why the updated versions of acs-packaging no longer do this. Happy to provide more info via zoom.

gavincornwell:
@alandavis Ah, yes, I remember that discussion but I thought the outcome was there wouldn't be any change. All the latest A and M builds are also on DockerHub so I guess I presumed nothing had changed!

@evasques In that case you'll also need to update the image location in the values file and docker compose file for 6.2.N. I've double-checked the deployment descriptor, the registry secret is already set for the repository pod so just changing the image tag and repository should be enough.

PR #461, opened 2020-11-20
OPSEXP-130: Share ingress setup + pin ingress-nginx version

sergiuv2020:
- added share ingress cookie
- edited docs and Travis to pin ingress-nginx chart version to 3.7.1
- added local charts to requirements to fix helm 3.41 linting


gavincornwell (approved):
LGTM

PR #463, opened 2020-11-23
OPSEXP-613: Revert artefacts to SP versions

gavincornwell:
We should NOT be using hotfix versions in our reference deployments according to our policy.

NOTE: There is no "6.0.1" tag for the repository available so I've updated to the latest hotfix for the 6.0.N deployment.


nilsSure (approved):
LG

PR #467, opened 2020-12-04
Fix typo in README

montgolfiere:


TheOutsSeeker (approved):
I guess, that's a no-brainer. :-)

gavincornwell (approved):
Thanks for catching that Jan.

PR #469, opened 2020-12-10
SEARCH-2506: Update Search Services to 2.0.1

aborroy:


gavincornwell (changes requested):
Can you please also update the following Helm chart values files:

- helm/alfresco-content-services/values.yaml
- helm/alfresco-content-services/community_values.yaml

aborroy:
Sure @gavincornwell 
The files have been updated.

gavincornwell:
Thanks @aborroy I'll keep an eye on the build and approve once green.

tpage-alfresco:
I can confirm that 2.0.1 is the correct version number.  It would be great to update fewer places when we make a release. In addition it seems strange that we're not testing with Insight Engine anywhere. Perhaps the enterprise files should use the insight-engine image rather than search-services?

tpage-alfresco (approved):

gavincornwell:
There is a lot less to update now than there used to be at least it's all in one repo now too ;-) But you're right it could be better, especially when we add Ansible playbooks to the mix too. We have discussed this within the team just had no time to do anything about it yet.

gavincornwell (approved):

PR #472, opened 2020-12-11
WIP: ACS-934: Add ooi service components to helm

KikoUK:
- Add both OOI Service and ADW env variables to the ACS Deployment helm charts (as an option)
- Document in Markdown as an Example extension and add link to new page within Deploy section


gavincornwell:
@KikoUK Can you please use a branch with a prefix of "feature/" so the Travis build executes (or add your branch name to the list in .travis.yml).

PR #475, opened 2020-12-14
Bump to ADW 2.0.0

montgolfiere:
- note: 2.0.0-adw (rather than 2.0.0)
- note: both master (7.0.0-Mx) *and* 6.2.N since new ADW / OOI extension targets ACS 6.2.2 or higher

### DIFF
--- a/docker-compose/6.2.N-docker-compose.yml
+++ b/docker-compose/6.2.N-docker-compose.yml
@@ -138,7 +138,7 @@ services:
             - 61613:61613 # STOMP
 
     digital-workspace:
-        image: quay.io/alfresco/alfresco-digital-workspace:1.6.0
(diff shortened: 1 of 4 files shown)


KikoUK (approved):
LGTM

gavincornwell:
@montgolfiere Why "2.0.0-adw" and not "2.0.0"? Is the OOI support only in "-adw"?

montgolfiere:
@gavincornwell It won't start properly. It seems 2.0.0 contains the APA code and the plugin is enabled by default (requested by PaaS). Hence as per @mauriziovitale it is better to use 2.0.0-adw

gavincornwell (approved):
LGTM

PR #480, opened 2020-12-17
OPSEXP-664: Allow custom repo debugging configuration to be provided for the ACS helm charts

nilsSure:
Created mechanism to allow custom repo debugging configuration to be provided for the ACS helm charts via values files.


gavincornwell (approved):
Looks good to me, as long as @sergiuv2020's concerns were addressed which it sounds like they were!

nilsSure:
> Looks good to me, as long as @sergiuv2020's concerns were addressed which it sounds like they were!

We had a call and checked together that it works as it is supposed to

PR #484, opened 2020-12-22
AAE-4349 Use activiti common chart

mteodori:


eromano (changes requested):
let's keep to use the FE chart

gavincornwell:
I agree with @eromano, let's stick with the ADF chart, unless there's a compelling reason to use the Activiti chart.

PR #483, opened 2020-12-18
AAE-4345 rename process to build tools

mteodori:

PR #485, opened 2021-01-04
OPSEXP-681: Build is failing on master

gavincornwell:
Just before Christmas several changes were made to the shared build scripts that have caused the ACS helm charts to start failing. As a temporary fix to unblock people I have disabled the checkov pre-commit hook for this repository and fixed the kubepug errors relating to ingress template files in the search chart.

https://github.com/Alfresco/acs-deployment/pull/483 is suffering with the same issue so we'll address the checkov failure in that PR.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,4 +7,4 @@ repos:
       - id: helm-lint
       - id: kubepug-minimum
       - id: kubepug-recommended
-      - id: checkov-helm
+      # - id: checkov-helm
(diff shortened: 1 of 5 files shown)


montgolfiere (approved):

montgolfiere:
@gavincornwell I've ticked but we still need green build so I assume further workaround changes are still needed ?

gavincornwell:
Yes, I cancelled the build as it only had the 6.0.N stage to go and had already proven the fixes had worked.

PR #486, opened 2021-01-04
[MNT-22131] Pull Share image from Docker Hub

sashcraft:
Customers do not have access to pull Share image from quay.io

### DIFF
--- a/docker-compose/docker-compose.yml
+++ b/docker-compose/docker-compose.yml
@@ -96,7 +96,7 @@ services:
             - shared-file-store-volume:/tmp/Alfresco/sfs
 
     share:
-        image: quay.io/alfresco/alfresco-share:7.0.0-M3
+        image: alfresco/alfresco-share:7.0.0-M3
         mem_limit: 1g
         environment:
             REPO_HOST: "alfresco"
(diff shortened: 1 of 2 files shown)


gavincornwell:
Hi Scott, you're absolutely right for the community deployments but the files you've changed are deploying the Enterprise version of ACS, there are several other images in the same file that come from quay.io, without access to quay.io the deployment will fail anyway. 

The prerequisites do state that customers need quay credentials: https://github.com/Alfresco/acs-deployment#prerequisites

Having said that, share is not an Enterprise product so should ideally be pulled from DockerHub (as the community files do). Initially the "M3" image was only available in quay.io hence the reference.

We have another issue on master at the moment causing builds to fail, I'll merge this once we have that resolved.

sashcraft:
This can be closed now that Share is available to customers from quay.io.

gavincornwell:
As discussed on Slack this should be resolved now, there was a label missing from some of the repositories in quay.io which prevented customers from seeing those images, this has hopefully been resolved now.

Also, FYI, due to the new [restrictions/limits](https://www.docker.com/blog/what-you-need-to-know-about-upcoming-docker-hub-rate-limiting/) on DockerHub we'll probably be making more use of quay.io and will default to quay.io for the Enterprise deployment scripts where access is required anyway.

gavincornwell:
Thanks for closing @sashcraft was literally just going to do the same!

PR #487, opened 2021-01-07
REPO-5203  Remove legacy transformer properties

alandavis:
With the removal of legacy transformers in ACS 7.0.0-A20, a number of properties have become obsolete.

Reviewers: Please take a special look at:
* config-repository.yml as the removal of lines in this file only make sense for the latest version of the content repository. If they are required for previous versions, we should undo these changes as they will be ignored by >= 7.0.0. [They were, so these changes were removed]
* README.md as I have just added an 5.0.0-M3 line without actually creating a new version. [also removed, as this is normally added when the release is done]


README.md
```diff
@@ -29,6 +29,7 @@ The table below shows the exact version of ACS deployed with each chart version/
 |--|--|--|--|--|--|
 | 5.0.0-M1 | 7.0.0-M2 | 6.2.2 | 6.1.1 | 6.0.1 | 6.2.1-A8 |
 | 5.0.0-M2 | 7.0.0-M3 | 6.2.2 | 6.1.1 | 6.0.1 | 7.0.0-A10 |
+| 5.0.0-M3 | 7.0.0-A20 | 6.2.2 | 6.1.1 | 6.0.1 | 7.0.0-A20 |
```
gavincornwell, line 32: We normally add this at the time of the release so it matches to the available tags.

docs/properties-reference.md
```diff
@@ -2176,8 +2178,8 @@ ldap.synchronization.defaultHomeFolderProvider | The default home folder provide
...
-| transform.misc.url | Legacy misc transformer url to T-Engines to service transform requests via http. Disabled by default. | |
+| ~~transform.misc.startupRetryPeriodSeconds~~ | ~~When the legacy misc transformer .url is set, this value indicates the amount of time to wait after a connection failure before retrying the connection to allow a docker container to (re)start.~~ | ~~60~~ |
+| ~~transform.misc.url~~ | ~~Legacy misc transformer url to T-Engines to service transform requests via http. Disabled by default.~~ | |
 | transform.service.cronExpression | Schedule for reading local transform config, so that T-Engines and local pipeline config is dynamically picked up, or reintegrated after an outage. Initially checks every 10 seconds and then switches to every hour after the configuration is read successfully. If there is a error later reading the config, the checks return to every 10 seconds. | `6 30 0/1 * * ?` |
 | transform.service.enabled | Determines whether the transform service is enabled | false |
 | transform.service.initialAndOnError.cronExpression | | `0/10 * * * * ?` |
```
gavincornwell, line 2185: Maybe we could add a NOTE just after the table to say that strikethrough'd properties have been removed in 7.0 (that will at least buy us a little time until we remove more in 7.1 ;-) )

Furthermore, the intention was for this list to move to the official docs where there are version specific pages so I think the note will be sufficient for now.

helm/alfresco-content-services/templates/config-repository.yaml
```diff
@@ -53,13 +53,6 @@ data:
       {{- if .Values.transformmisc.enabled }}
       -DlocalTransform.misc.url=http://{{ template "content-services.shortname" . }}-transform-misc
       {{- end }}
-      -Dalfresco-pdf-renderer.url=http://{{ template "content-services.shortname" . }}-pdfrenderer
```
gavincornwell, line 56: This file is shared across all deployments so we'll have to find a way to optionally include/exclude these properties based on repo version.

For now, if they cause no harm being there in 7.0 let's just leave them there for now, but please add a comment to remind us to address it.

gavincornwell (changes requested):

README.md
alandavis: Removed line

helm/alfresco-content-services/templates/config-repository.yaml
alandavis: Agreed. Restored

docs/properties-reference.md
alandavis: * There already is a note at the top of the table about strikethrough text. Originally it was at the end, but as it is such a long page, it is not seen.

* I was talking with on Adelaide about transform properties on Monday (11 Jan), and pointed her at this list. I don't think they have been included yet in the official docs.

gavincornwell (approved):
LGTM

PR #492, opened 2021-02-10
OPSEXP-755 & OPSEXP-756: Update nginx config

gavincornwell:
Updated to new version of alfresco-acs-nginx that now supports Enterprise and Community and includes the required change.

Added additional tests for the extra protected URLs.


nilsSure (approved):
LGTM

PR #495, opened 2021-02-19
ATS-862: Release ATS 1.3.2-A2

TheOutsSeeker:


docker-compose/6.1.N-docker-compose.yml
```diff
@@ -47,7 +47,7 @@ services:
 
     transform-router:
         mem_limit: 512m
-        image: quay.io/alfresco/alfresco-transform-router:1.0.2
+        image: quay.io/alfresco/alfresco-transform-router:1.3.2-A2
```
montgolfiere, line 50: ACS 6.1.N  might be one step back too far unless explicitly tested for backwards-compatibility.

I'd start with at least master / ACS 7.0.0 for now (docker-compose.yml) and potentially ACS 6.2.N although should be sanity checked.

docker-compose/6.2.N-docker-compose.yml
```diff
@@ -48,7 +48,7 @@ services:
 
     transform-router:
         mem_limit: 512m
-        image: quay.io/alfresco/alfresco-transform-router:1.3.1
```
montgolfiere, line 51: See previous comment.

montgolfiere (changes requested):

docker-compose/6.1.N-docker-compose.yml
TheOutsSeeker: so, what's your suggestion - revert the change?

PR #497, opened 2021-02-23
OPSEXP-784: fix dockerhub setup

sergiuv2020:


.travis.yml
```diff
@@ -206,7 +215,6 @@ _helm_deploy: &helm_deploy
     --set global.alfrescoRegistryPullSecrets=quay-registry-secret \
     --wait \
     --timeout 20m0s \
-    --atomic \
```
gavincornwell, line 209: I know you removed this for testing but did you intend to leave it out of forget to add back?

.travis.yml
```diff
@@ -190,7 +200,6 @@ _helm_deploy: &helm_deploy
     --set controller.service.annotations."service\.beta\.kubernetes\.io/aws-load-balancer-security-groups"="${AWS_SG}" \
     --set controller.publishService.enabled=true \
     --wait \
-    --atomic \
```
gavincornwell, line 193: I know you removed this for testing but did you intend to leave it out of forget to add back?

.travis.yml
```diff
@@ -260,7 +267,6 @@ _helm_deploy: &helm_deploy
         helm upgrade $release_name_acs helm/alfresco-content-services \
         --wait \
         --timeout 10m0s \
-        --atomic \
```
gavincornwell, line 263: I know you removed this for testing but did you intend to leave it out of forget to add back?

.travis.yml
sergiuv2020: the cleanup is done at the end anyway, sadly with this atomic if the deployment fails initially it will get deleted on the spot even if keep env is set, so its better to remove it actually.

.travis.yml
sergiuv2020: removed on purpose

.travis.yml
sergiuv2020: removed on purpose

gavincornwell (approved):
Thanks for the clarification

PR #500, opened 2021-02-24
Feature/update sync service

ciprianchelaru:
Update Sync Service from 3.4.0-M2 to 3.4.0-M5. This update is only for ACS 7.0.0


gavincornwell (approved):
LGTM

PR #501, opened 2021-02-24
OPSEXP-105: Minor tweak to Community Docker Compose diagram

montgolfiere:
- Core AIO does not connect to ActiveMQ (for Community ref deployment)
- eg. see

https://github.com/Alfresco/acs-deployment/blob/master/docker-compose/community-docker-compose.yml#L48-L54
https://github.com/Alfresco/acs-deployment/blob/master/docker-compose/docker-compose.yml#L69-L79


gparapadakis (approved):

montgolfiere:
@gavincornwell Noticed this after chatting with @gparapadakis.

As an aside (maybe a separate PR) we could also look to tweak the Enterprise diagrams for both docker-compose and helm. The T-Engines also call ActiveMQ. However, I'm not sure how well it would display for the helm/k8s option with so many crossing lines.

gavincornwell (commented):
Has the image generated correctly, shows as empty!?

montgolfiere:
I think so. I can download it :-) 

Maybe a GitHub web ui issue ?

If you click on rich diff you can see Before and After.

gavincornwell:
Hmm, you're right, is showing correctly now!?

gavincornwell (approved):

PR #502, opened 2021-02-24
ACS-1183 ACS 7 Stacks: PostgreSQL

alandavis:
Upgraded to PostgreSQL 13.1 in ACS 7.0.0

### DIFF
--- a/helm/alfresco-content-services/values.yaml
+++ b/helm/alfresco-content-services/values.yaml
@@ -509,7 +509,7 @@ postgresql:
   # Note: Set this to false if you use an external database.
   enabled: true
   nameOverride: postgresql-acs
-  imageTag: "11.7.0"
+  imageTag: "13.1"
   commonAnnotations:
     application: alfresco-content-services
   postgresqlUsername: alfresco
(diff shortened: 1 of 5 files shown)


helm/alfresco-content-services/values.yaml
```diff
@@ -509,7 +509,7 @@ postgresql:
   # Note: Set this to false if you use an external database.
   enabled: true
   nameOverride: postgresql-acs
-  imageTag: "11.7.0"
+  imageTag: "13.1"
   commonAnnotations:
```
alandavis, line 513: Not sure this is correct, as I would not have expected 11.7.0 as the original value.

helm/alfresco-content-services/values.yaml
```diff
@@ -534,7 +534,7 @@ postgresql-syncservice:
   enabled: true
   replicaCount: 1
   nameOverride: postgresql-syncservice
-  imageTag: "11.7.0"
+  imageTag: "13.1"
   commonAnnotations:
```
alandavis, line 538: Not sure this is correct, as I would not have expected 11.7.0 as the original value. Community used just 11.7.

gavincornwell:
@alandavis Do you mind if I close this one as I've done the same change (and more) in this branch: https://github.com/Alfresco/acs-deployment/tree/feature/upgrade-postgres-share

NOTE: The Helm charts use the Bitnami Postgres chart which in turn use their own Postgres images hence the slightly different format.

gavincornwell (approved):
LGTM.

NOTE: The Helm image tag looks slightly different as we use the Bitnami Postgres Helm chart which in turn uses their own Postgres image.

Issue #490, opened 2021-02-09
Read timeout error on search scheduler worker while starting up containers


abhinavmishra14:
Read timeout error on search scheduler worker while starting up containers.

Used this latest version of docker-compose: https://github.com/Alfresco/acs-deployment/blob/master/docker-compose/community-docker-compose.yml

`alfresco_1            | 2021-02-09 18:10:02,061  ERROR [quartz.core.JobRunShell] [SearchScheduler_Worker-1] Job Solr.SolrWatcher threw an unhandled Exception:
alfresco_1            | org.alfresco.error.AlfrescoRuntimeException: 01090038 Read timed out
alfresco_1            |         at org.alfresco.repo.search.impl.solr.AbstractSolrAdminHTTPClient.getOperation(AbstractSolrAdminHTTPClient.java:96)
alfresco_1            |         at org.alfresco.repo.solr.SOLRAdminClient.executeAction(SOLRAdminClient.java:221)
alfresco_1            |         at org.alfresco.repo.solr.SOLRAdminClient$SolrTracker.pingSolr(SOLRAdminClient.java:333)
alfresco_1            |         at org.alfresco.repo.solr.SOLRWatcherJob.execute(SOLRWatcherJob.java:54)
alfresco_1            |         at org.quartz.core.JobRunShell.run(JobRunShell.java:202)
alfresco_1            |         at org.quartz.simpl.SimpleThreadPool$WorkerThread.run(SimpleThreadPool.java:573)
alfresco_1            | Caused by: java.net.SocketTimeoutException: Read timed out
alfresco_1            |         at java.base/java.net.SocketInputStream.socketRead0(Native Method)
alfresco_1            |         at java.base/java.net.SocketInputStream.socketRead(SocketInputStream.java:115)
alfresco_1            |         at java.base/java.net.SocketInputStream.read(SocketInputStream.java:168)
alfresco_1            |         at java.base/java.net.SocketInputStream.read(SocketInputStream.java:140)
alfresco_1            |         at java.base/java.io.BufferedInputStream.fill(BufferedInputStream.java:252)
alfresco_1            |         at java.base/java.io.BufferedInputStream.read(BufferedInputStream.java:271)
alfresco_1            |         at org.apache.commons.httpclient.HttpParser.readRawLine(HttpParser.java:78)
alfresco_1            |         at org.apache.commons.httpclient.HttpParser.readLine(HttpParser.java:106)
[...]

PR #510, opened 2021-03-02
OPSEXP-805: Increased connection timeout for repo to search communication

gavincornwell:
Fix for https://github.com/Alfresco/acs-deployment/issues/490


sergiuv2020:
Can you also change the steps for waiting on activemq async. 300 it's a bit to much.

lsuciu (approved):

gavincornwell:
@sergiuv2020 Yes, we could reduce the activemq timeout, however, we use this setting everywhere and that's not the purpose of this PR so I'd suggest we raise that separately.

PR #520, opened 2021-03-11
OPSEXP-827: ACS Helm deployment is still using Postgres 11 despite being configured with 13

nilsSure:
Replaced _imageTag_ with image.tag as referred in bitnami charts. Also, 13.1 tag doesn't exists in bitnami's docker repository, but they do have 13.1.0


gavincornwell (approved):
LGTM

PR #535, opened 2021-04-01
ATS-884: Attempting to make t-router work on terraform pipeline

aymanthefirst:


aymanthefirst:
Renamed the branch so that it is picked up by tests

PR #544, opened 2021-04-08
OPSEXP-435: Automatic chart docs with pre-commit and helm-docs

sergiuv2020:


gavincornwell (changes requested):
If we do this we should replace the contents of the "Configure" section (https://github.com/Alfresco/acs-deployment/tree/master/docs/helm#configure) and point to the new page. 

However, doing that we'll lose the descriptions, is there a way to specify them with the helm-docs tool?

sergiuv2020:
I will update the values files so we can still get those descriptions

gavincornwell (approved):
LGTM

PR #545, opened 2021-04-08
Fix pre-commit hook failure

gavincornwell:
Run pre-commit hook to include new properties from values files.

NOTE: Had to add a blank line to search values file to avoid incorrect docs generation.


nilsSure (approved):
LGTM

PR #551, opened 2021-04-29
ATS-892: IPTC model

montgolfiere:
- note: WIP - see also T-Core (alfresco-transform-core)
-
- example added to both Enterprise & Community


skopf (changes requested):
See also conversation in Jira: Why can't we add these models to the repository just like the EXIF and DublinCore models?

montgolfiere:
@skopf Yes, agreed. This was a temporary PR to demonstrate how it could be done via docker-compose when using ACS 7.0.0 &/or 6.2.2.

Closing the Draft PR without merge, since the IPTC Model (& related message files) are now part of core Repo (on "main"):

- https://github.com/Alfresco/alfresco-community-repo/blob/master/repository/src/main/resources/alfresco/model/iptcModel.xml
- https://github.com/Alfresco/alfresco-community-repo/tree/master/repository/src/main/resources/alfresco/messages

PR #553, opened 2021-05-07
Update eks-deployment.md

john-knowles:
Suggest update the apiVersion in the sample external-dns.yaml from rbac.authorization.k8s.io/v1beta1 to rbac.authorization.k8s.io/v1


john-knowles:
migrated these changes to pull request #554.

PR #554, opened 2021-05-07
Update eks-deployment.md

john-knowles:
Suggest update the apiVersion in the sample external-dns.yaml from rbac.authorization.k8s.io/v1beta1 to rbac.authorization.k8s.io/v1

And suggestion for additional step in **Prepare The Cluster For ACS** where the nodegroup name created by following the previous steps is not _linux-nodes_ and causes the command to fail (previously step 4).


mteodori (approved):
LGTM

PR #567, opened 2021-06-16
Opsexp 922 chown for volumes

bartek-lopatka:
Ref: [OPSEXP-922](https://alfresco.atlassian.net/browse/OPSEXP-922)

Got rid of init-fs containers, because we can't run `chown` command with  EFS CSI dynamic provisioning ([bug](https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/300) is already created for provider).

### DIFF
--- a/helm/alfresco-content-services/charts/activemq/templates/deployment-activemq.yaml
+++ b/helm/alfresco-content-services/charts/activemq/templates/deployment-activemq.yaml
@@ -64,6 +64,7 @@ spec:
           subPath: {{ .Values.persistence.subPath }}
         resources:
 {{ toYaml .Values.resources | indent 12 }}
+      {{- if .Values.persistence.dynamicProvisioning }}
       initContainers:
         - name: init-fs
           image: "{{ .Values.initContainer.image.repository }}:{{ .Values.initContainer.image.tag }}"
@@ -80,4 +81,5 @@ spec:
       - name: data
         persistentVolumeClaim:
           claimName: {{ .Values.persistence.existingClaim }}
+      {{- end }}
 {{- end }}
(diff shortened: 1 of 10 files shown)


helm/alfresco-content-services/templates/deployment-repository.yaml
```diff
@@ -82,9 +82,11 @@ spec:
             mountPath: /usr/local/tomcat/shared/classes/alfresco/extension/dev-log4j.properties
             subPath: dev-log4j.properties
           {{- end }}
+          {{- if .Values.persistence.dynamicProvisioning }}
```
alxgomz, line 85: I would argue we would need volume to be mounted even if not using dynamic provisioning

helm/alfresco-content-services/charts/activemq/templates/deployment-activemq.yaml
```diff
@@ -58,12 +58,13 @@ spec:
           periodSeconds: {{ .Values.livenessProbe.periodSeconds }}
           failureThreshold: {{ .Values.livenessProbe.failureThreshold }}
           timeoutSeconds: {{ .Values.livenessProbe.timeoutSeconds }}
+        resources:
+{{ toYaml .Values.resources | indent 12 }}
+       {{- if .Values.persistence.dynamicProvisioning }}
```
alxgomz, line 63: I would argue we would need volume to be mounted even if not using dynamic provisioning

helm/alfresco-content-services/charts/alfresco-search/templates/deployment.yaml
```diff
@@ -51,10 +51,12 @@ spec:
             - containerPort: {{ template "alfresco-search.internalPort" . }}
           resources:
 {{ toYaml .Values.resources | indent 12 }}
+          {{- if .Values.persistence.dynamicProvisioning }}
```
alxgomz, line 54: same as above I think

mteodori:
in order to build properly prefix the branch with fix/
also remove gp2 as storageclass in order not to have anything AWS specific

PR #573, opened 2021-07-09
SEARCH-2878 Update the supported SearchService version

jesty:


mteodori (approved):

mteodori:
@jesty @killerboot should also the 7.0.N values and docker compose be updated?

mteodori:
I will fix directly on master the missing changes and the pre-commit as the branch name didn't match feature/* to trigger the check

PR #580, opened 2021-08-12
DESKTOPAPP-721: Bump Sync service from 3.4.0 to 3.4.1-A1

ssingh09:


ciprianchelaru (approved):
LGTM but see how helm chart version should be updated

mteodori:
> LGTM but see how helm chart version should be updated

when releasing Burnham we will update chart version and release 5.1.0-M3 no need to do anything in this PR

mteodori (approved):

PR #583, opened 2021-08-25
ACS-1870: Add MS Teams option to ACS Helm charts (part 2)

montgolfiere:
- make sure alfresco.base-url (eg. https://my-env:443) is also configurable
- used to generate ADW preview links (in addition to Alfresco REST API calls)


mpichura (approved):

aymanthefirst (approved):
Looks good to me as long as the variables are passing through

montgolfiere:
@mpichura @aymanthefirst Thanks.

@aymanthefirst JFYI: I've also sanity checked using the documented example that values are passed through in a local k8s cluster - this can be seen in the startup log of the ms-team-service.

PR #595, opened 2021-09-01
ACS-1914: Update helm charts for MS Teams example

montgolfiere:
- add new property "alfresco.digital-workspace.base-url" (introduced in 0.5.0-A1)

- see also ACS-1870, ACS-1908

### DIFF
--- a/docs/helm/examples/with-ms-teams.md
+++ b/docs/helm/examples/with-ms-teams.md
@@ -41,7 +41,8 @@ helm install acs alfresco/alfresco-content-services \
 --set messageBroker.user="alfresco" \
 --set messageBroker.password="YOUR-MQ-PASSWORD" \
 --set msTeams.enabled=true \
---set msTeamsService.alfresco.baseUrl="https://acs.YOUR-DOMAIN-NAME:443"
+--set msTeamsService.alfresco.baseUrl="https://acs.YOUR-DOMAIN-NAME:443" \
+--set msTeamsService.alfresco.digitalWorkspace.baseUrl="https://acs.YOUR-DOMAIN-NAME:443" \
 --set msTeamsService.alfresco.digitalWorkspace.contextPath="/workspace/" \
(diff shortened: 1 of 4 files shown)


docs/helm/examples/with-ms-teams.md
```diff
@@ -41,7 +41,8 @@ helm install acs alfresco/alfresco-content-services \
 --set messageBroker.user="alfresco" \
 --set messageBroker.password="YOUR-MQ-PASSWORD" \
 --set msTeams.enabled=true \
---set msTeamsService.alfresco.baseUrl="https://acs.YOUR-DOMAIN-NAME:443"
+--set msTeamsService.alfresco.baseUrl="https://acs.YOUR-DOMAIN-NAME:443" \
```
mpichura, line 44: Didn't we want to put `http://localhost:8080` in `alfresco.baseUrl` when thought about introducing `alfresco.digitalWorkspace.baseUrl`?

mpichura (approved):

montgolfiere:
As per Team decision, we may revert the new property for now and release 0.6.0-A1 (or even 1.0.0-A1).  Hence, marking as Draft (and we may close without merge).

docs/helm/examples/with-ms-teams.md
montgolfiere: Yes. I think here, it could be the default Kubernetes internal url, eg.

https://github.com/Alfresco/acs-deployment/blob/master/helm/alfresco-content-services/values.yaml#L119

`http://acs-alfresco-cs-repository:80`

However, as per [team review](https://github.com/Alfresco/ms-teams-integration/pull/43) (private repo) we'd need both internal and external urls for different usages.

montgolfiere:
No longer required (for 1.0.0-A1) after https://github.com/Alfresco/ms-teams-integration/pull/45

Hence, closing without merge.

PR #596, opened 2021-09-03
Feature/desktopapp 724

ssingh09:
Bump sync service from 3.4.1-A1 to 3.4.1-M1


yuvarajpandurangan (approved):
Looks good to me, so approving it.

PR #600, opened 2021-09-06
ATS-955 Bump AIS -> 1.4.0-A1

PunchedChimera:


montgolfiere (changes requested):
I suggest we only bump AIS version here (not Repo/Share) ...

montgolfiere (approved):

PR #602, opened 2021-09-14
ATS-964 Bump ATS 1.4.1

PunchedChimera:
Includes:
- ATS-967 SFS to 0.16.0
- ATS-966 AIS to 1.4.0
- ACS-1967 OOI to 1.1.1


montgolfiere (approved):
@PunchedChimera I realise it is still marked as Draft but it seems ok to me 👍

PR #611, opened 2021-10-06
OPSEXP-1095 csi driver update

alxgomz:
Use an init container in order to make sure postgres is happy with data dir permissions


helm/alfresco-content-services/7.0.N_values.yaml
```diff
@@ -63,3 +63,7 @@ alfresco-digital-workspace:
   image:
     repository: quay.io/alfresco/alfresco-digital-workspace
     tag: 2.4.0-adw
+
+postgresql:
+  persistence:
+    storageClass: "gp2"
```
mteodori, line 69: do we need this? gp2 is AWS only and the EKS default

TheOutsSeeker (approved):

PR #618, opened 2021-10-21
OPSEXP-1077 update download-trial branch to ACS 7.1

alxgomz:
update download-trial to serve 7.1 files

**Note: Not sure this is the proper way to release that new version**

Ref OPSEXP-1077

### DIFF
--- a/docker-compose/community-docker-compose.yml
+++ b/docker-compose/community-docker-compose.yml
@@ -10,7 +10,7 @@ version: "2"
 
 services:
     alfresco:
-        image: alfresco/alfresco-content-repository-community:7.1.0
+        image: alfresco/alfresco-content-repository-community:7.1.0.1
         mem_limit: 1500m
         environment:
             JAVA_TOOL_OPTIONS: "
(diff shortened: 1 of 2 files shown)


mteodori (commented):
I don't see the need and value in having this branch build as it should already have been tested when tagging master with v5.1.0

when confirmed we can publish we can just push force the v5.1.0 tag on this dummy branch and close this PR

alxgomz:
> I don't see the need and value in having this branch build as it should already have been tested when tagging master with v5.1.0
> 
> when confirmed we can publish we can just push force the v5.1.0 tag on this dummy branch and close this PR

TBH, I'm not sure either. Maybe having a separate branch is a choice made in order to coordinate download-trial release with online-trial release. Other wise if the download manager were t o use paster and acs-deployemnt is released but not online-trial (as it happens almost on every release), then the online & download trail would serve different versions of the product which would be confusing.

As to why this is a branch and not a tag... The only reason I can think of is that it would allow to divert from master if needed at some point....?
I *think* moving to a tag instead would not break the download manager assuming it's just using a simple URL pattern... But that's just me guessing. 
I sent @rickyclemmons an email about that to see if he can find any information about it.

PR #624, opened 2021-11-17
AAE-5415: WIP (Do Not Merge) - example to run Admin App with ACS

montgolfiere:
- note: do not merge as-is (in future would need to test with Alfresco proxy image - ie. to keep as single docker-compose file)


mteodori:
this PR is very useful as it removes the need for the custom alfresco nginx image so we could definitely retire that project

I would avoid having a new docker-compose config though to prevent duplication but sadly docker-compose doesn't have any concept of import or conditional so I can only think of a separate file with just AAA

montgolfiere:
Note: this will be superseded by https://alfresco.atlassian.net/browse/OPSEXP-1226

PR #648, opened 2022-02-01
OPSEXP-1163: Support shared secret for REPO/SOLR communication

EquinoxBlack:


skopf:
The secret key needs to either be configurable or be auto-generated as a random string, rather than just using the fixed static string "secret"

EquinoxBlack:
> The secret key needs to either be configurable or be auto-generated as a random string, rather than just using the fixed static string "secret"

@In this case could this be specified in the ticket please? Also currently I am struggling to deploy acs with the config as inn this pr - the solr pod is not getting up..

helm/alfresco-content-services/README.md
```diff
@@ -97,10 +97,12 @@ Hence, setting up explicit Container memory and then assigning a percentage of i
...
+| global.alfrescoRegistryPullSecrets | string | `nil` |  |
 | global.registryPullSecrets[0] | string | `"quay-registry-secret"` |  |
 | global.strategy.rollingUpdate.maxSurge | int | `1` |  |
 | global.strategy.rollingUpdate.maxUnavailable | int | `0` |  |
+| global.tracking.auth | string | `"secret"` | Select how solr and repo authenticate to each other none: work only prior to acs 7.2 (and was the default) secret: use a shared secret (to specify using `tracking.sharedsecret`) https: to use mTLS auth (require appropriate certificate configuration) |
+| global.tracking.sharedsecret | string | `"F7nRvqh0jKJnGlhzLKsdQiZ3/37x6Yb7f8N8i0HE52Q"` | Shared secret to authenticate repo/solr traffic |
```
gionn, line 105: there are any minimum requirements for this secret?
suggesting to use something like `openssl rand -hex 20` may strength security

skopf (changes requested):
I have concerns that the shared secret defaults to the string "secret" if not provided:
```
{{ .Values.global.tracking.sharedsecret | default "secret" }}
```
Instead, the helm chart should fail.
At least, we should remove this default value so that the container startup fails due to a missing secret. Ideally, the HELM charts would fail early and not try to setup anything.

helm/alfresco-content-services/charts/alfresco-search/README.md
```diff
@@ -37,7 +38,7 @@ A Helm chart for deploying Alfresco Search
 | insightEngineImage.internalPort | int | `8983` |  |
 | insightEngineImage.pullPolicy | string | `"IfNotPresent"` |  |
 | insightEngineImage.repository | string | `"quay.io/alfresco/insight-engine"` |  |
-| insightEngineImage.tag | string | `"2.0.0"` |  |
+| insightEngineImage.tag | string | `"2.1.0-SNAPSHOT"` |  |
```
alandavis, line 41: You are probably correct, but I thought it was 2.0.3 rather than 2.1.0 https://alfresco.atlassian.net/browse/ACS-2305 - Going to the next minor makes sense for this change.

skopf (changes requested):
Again, why do we have a default value for the shared secret? We either need to force the user to provide one themselves or we need to create a random one.
But we should never provide a default value, even if it is F7Rvqh.j&:nlzLsdQi3/37xYb$f8N80HE52Q
The result will only be that we have a ton of systems running out there in the wild using this value as shared secret.

Can we please remove the default value?

helm/alfresco-content-services/charts/alfresco-search/README.md
dsibilio: The changes in this branch needed the Docker image support for `ALFRESCO_SECURE_COMMS=secret`, which was originally only available on the latest `2.1.0-SNAPSHOT` build. Since the [IE/SS release build](https://app.travis-ci.com/github/Alfresco/InsightEngine/jobs/561786320) finally ran yesterday, I think it'd be safe to test with `2.0.3-RC2`, which should include the aforementioned Docker image support.

skopf (approved):
Nice work!

jamalkm:
Any update on this? Will it be merge soon?

jamalkm (approved):

.pre-commit-hooks.yaml
```diff
@@ -1,5 +1,5 @@
 - id: checkov-helm
   name: Checkov - Helm check
-  entry: bash -c "helm template helm/alfresco-content-services --set \"alfresco-search.alfresco-insight-zeppelin.insightzeppelin.enabled\"=true > helm/alfresco-content-services.yaml && checkov -f helm/alfresco-content-services.yaml --framework=kubernetes --skip-check $(cat .checkov-ignore-list) --quiet"
+  entry: checkov -d helm/alfresco-content-services --framework helm --var-file .checkov/helm_vars.yaml --skip-check $(cat .checkov/skip-list) --quiet
```
gionn, line 3: I think you can get rid of this file since it's not really used anywhere else https://github.com/search?l=YAML&q=org%3AAlfresco+checkov-helm&type=Code

PR #652, opened 2022-02-09
add helper to select the good api version, change rules shape for eac…

TOTOleHero:
Fix to manage ingress API version.
Change ingress rules vs API version


mteodori:
superceded by #672

PR #656, opened 2022-02-18
add port to share csrf filter configuration

hlavki:


mteodori:
Thanks @hlavki can you add an explanation to the PR on why would we need this change? Is there anything broken without the port in this configuration setting we can test if it's fixed once we merge?

hlavki:
Hi @mteodori, you are out of luck, if you need to access `/share` application on different port as default ports (80, 443).

PR #661, opened 2022-03-02
OPSEXP-1035 Updated ActiveMQ to 5.16.4

mteodori:
Ref. OPSEXP-1035

### DIFF
--- a/docker-compose/6.1.N-docker-compose.yml
+++ b/docker-compose/6.1.N-docker-compose.yml
@@ -163,7 +163,7 @@ services:
       - "8083:8983" # Browser port
 
   activemq:
-    image: jnix85/activemq:5.16.4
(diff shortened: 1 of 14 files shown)


alxgomz (commented):
I'm worried about the fact that image comes from yet another sources which seems to be using a customized openjdk. Note I don't know the details of their customization so it could even be better than the vanilla one... But I' must highlighting that we (or at least I) know so few about it.

mteodori:
> I'm worried about the fact that image comes from yet another sources which seems to be using a customized openjdk. Note I don't know the details of their customization so it could even be better than the vanilla one... But I' must highlighting that we (or at least I) know so few about it.

reverted to our updated ActiveMQ until the official image is available with https://issues.apache.org/jira/browse/AMQ-8149

PR #666, opened 2022-03-10
OPSEXP-1318 Fix YAML parse error when sharedsecret contains special characters

gionn:
Fix for:
```
synchronization of release 'acs' in namespace 'acs' failed: installation failed: YAML parse error on alfresco-content-services/charts/alfresco-search/templates/deployment.yaml: error converting YAML to JSON: yaml: line 60: did not find expected whitespace or line break
```

when a sharedsecret is provided with special YAML characters (e.g.: leading `!`)

OPSEXP-1318


helm/alfresco-content-services/charts/alfresco-search/values.yaml
```diff
@@ -99,7 +99,7 @@ global:
     # https: to use mTLS auth (require appropriate certificate configuration)
     auth: secret
     # -- Shared secret to authenticate repo/solr traffic
-    #sharedsecret:
+    sharedsecret: null
```
gionn, line 102: @alxgomz why this was commented?

alxgomz: No good reason... It's not even consistent with the main value file.

alxgomz (approved):
LGTM

PR #674, opened 2022-03-18
OPSEXP-1199 Adds and documents a license installation mechanism

alxgomz:
introduces the value `repository.licenseSecret` to point to the kubernetes secret where a license has previously been deployed (choice made to stick to `RegistryPullSecret`).
Doing so it is possible to deploy from start an ACS environment with its license (without uploading through the UI), and more importantly the renewal of the license can be done on the fly and triggers pods' updates so it is taken into account immediately


docs/helm/examples/alf_license.md
```diff
@@ -0,0 +1,44 @@
+# Deploy Alfresco Content Services with the ability to manage licenses
+
+Alfresco content repository Docker images come with an embeded license which last only 2 days. If you've built a custom image, it may have a one year license but in order to change it you would need to biuld a new image.
```
gionn, line 3:
```suggestion
Alfresco content repository Docker images come with an embedded license which last only 2 days. If you've built a custom image, it may have a one year license but in order to change it you would need to build a new image.
```

PR #684, opened 2022-04-01
OPSEXP-1374: Fix ooi and ms teams service - http 404 error

Mieszko96:
Ref. OPSEXP-1374


MichalFidor (approved):

montgolfiere (approved):

montgolfiere (changes requested):
If it works then LGTM.

Same may apply to other (optional) external API endpoints, eg. /ms-teams-service ?

https://github.com/Alfresco/acs-deployment/blob/master/helm/alfresco-content-services/templates/ingress-ms-teams-service.yaml#L12

alxgomz (approved):
LGTM

montgolfiere (approved):
Thanks

mteodori (approved):
LGTM, those annotations seem incorrect anyway as they are not configurable while ingress.path is

PR #693, opened 2022-04-13
OPSEXP-1369 Add elasticsearch connector

gionn:
OPSEXP-1369


helm/alfresco-content-services/templates/config-repository.yaml
```diff
@@ -40,13 +40,19 @@ data:
       {{- if index .Values "alfresco-search" "enabled" }}
       -Dsolr.host={{ template "alfresco-search.host" . }}
       -Dsolr.port={{ template "alfresco-search.port" . }}
+      -Dsolr.secureComms={{ .Values.global.tracking.auth | default "none" }}
```
alxgomz, line 43: It was explicitly requested by architects (@skopf) that the default is set to `secret` in case search2 is enabled.

helm/alfresco-content-services/templates/secret-database.yaml
```diff
@@ -1,10 +1,10 @@
...
-  name: {{ template "content-services.shortname" . }}-dbsecret
+  name: {{ .Values.global.repo.databaseSecretName }}
   labels:
     {{- include "repository.labels" . | nindent 4 }}
 type: Opaque
 data:
-  DATABASE_USERNAME: {{ .Values.database.user | default .Values.postgresql.postgresqlUsername | b64enc | quote }}
-  DATABASE_PASSWORD: {{ .Values.database.password | default .Values.postgresql.postgresqlPassword | b64enc | quote }}
+  {{ .Values.global.repo.databaseSecretUsernameKey }}: {{ .Values.database.user | default .Values.postgresql.postgresqlUsername | b64enc | quote }}
+  {{ .Values.global.repo.databaseSecretPasswordKey }}: {{ .Values.database.password | default .Values.postgresql.postgresqlPassword | b64enc | quote }}
```
alxgomz, line 10: We can't leave the option to override the pg credentials in the `database` context otherwise there's no way for the subchart to authenticate to the db.
Also I don't see the benefit of using input values as keys neither for secret name. It just makes things a lot more complex to configure.
I'd rather we instead use static values and reference those in the subchart

helm/alfresco-content-services/values.yaml
```diff
@@ -859,6 +870,11 @@ global:
   ai:
     enabled: false
 
+  repo:
+    databaseSecretName: acs-alfresco-cs-dbsecret
+    databaseSecretUsernameKey: DATABASE_USERNAME
+    databaseSecretPasswordKey: DATABASE_PASSWORD
+
```
alxgomz, line 877: Has explained above I don't think any of these should be required.

helm/alfresco-content-services/charts/alfresco-elasticsearch-connector/templates/liveindexing-config.yaml
```diff
@@ -0,0 +1,23 @@
...
+metadata:
+  name: {{ template "alfresco-elasticsearch-connector.fullName" . }}-configmap
+  labels:
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    component: search
+data:
+  ELASTICSEARCH_INDEXNAME: "{{ .Values.indexName }}"
+  SPRING_ELASTICSEARCH_REST_URIS: http://elasticsearch-master:9200
```
alxgomz, line 12: This might need to be set to some other value, or we need an external service resource matching that name.

helm/alfresco-content-services/charts/alfresco-elasticsearch-connector/templates/reindexing-config.yaml
```diff
@@ -0,0 +1,22 @@
...
+metadata:
+  name: {{ template "alfresco-elasticsearch-connector.fullName" . }}-reindexing-configmap
+  labels:
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    component: search
+data:
+  ELASTICSEARCH_INDEXNAME: "{{ .Values.indexName }}"
+  SPRING_ELASTICSEARCH_REST_URIS: http://elasticsearch-master:9200
```
alxgomz, line 12: This might need to be set to some other value, or we need an external service resource matching that name.

alxgomz (changes requested):

helm/alfresco-content-services/templates/config-repository.yaml
gionn: that is true now, because before was set regardless of search2 enabled or not (or maybe those ifs are very hard to read for me?)

helm/alfresco-content-services/templates/config-repository.yaml
```diff
@@ -40,13 +40,19 @@ data:
...
       {{- else }}
       {{- if index .Values "alfresco-search" "external" }}
       -Dsolr.host={{ index .Values "alfresco-search" "external" "host" }}
       -Dsolr.port={{ index .Values "alfresco-search" "external" "port" }}
+      -Dsolr.secureComms={{ .Values.global.tracking.auth | default "none" }}
       {{- end }}
       {{- end }}
-      -Dsolr.secureComms={{ .Values.global.tracking.auth | default "none" }}
+      {{- if index .Values "alfresco-elasticsearch-connector" "enabled" }}
```
gionn, line 51: do I need here to start using standard camelCase naming convention and avoid using lowercase-separated-with-dash?

helm/alfresco-content-services/templates/secret-database.yaml
gionn: those values are defined to avoid repeating static strings here and in the subchart, but I would argue that for the user having Values that are not meant to be changed are just confusing?

I can just use `.Release.name` in metadata.name here and in the subchart, and DATABASE_* as static strings

helm/alfresco-content-services/templates/config-repository.yaml
alxgomz: yes, the if logic works best like this but the default is still `none`. But tbh that was also the case before... up to you.

helm/alfresco-content-services/templates/config-repository.yaml
alxgomz: We might regret it at some point but let's stick to dash notation for now...

helm/alfresco-content-services/templates/secret-database.yaml
alxgomz: Can't we proceed as if the subchart is primarily meant to be used with the acs-deployment chart (and so use  "content-services.shortname" to expand the secret name just like in the main chart) and default to something else in case that template is not avaialble (e.g. an input value which would only live in the subchart values)?

.pre-commit-config.yaml
```diff
@@ -25,14 +24,9 @@ repos:
...
-      - id: check-travis
-  - repo: local
-    hooks:
-      - id: checkov-helm
-        name: Checkov - Helm check
-        entry: bash -c "checkov -d helm/alfresco-content-services --framework helm --var-file .checkov/helm_vars.yaml --skip-check $(cat .checkov/skip-list) --quiet"
-        language: system
-        pass_filenames: false
+    - id: checkov
+      args: [--config, .checkov/config.yml]
```
alxgomz, line 31: 🥇

helm/alfresco-content-services/charts/alfresco-elasticsearch-connector/templates/reindexing-config.yaml
```diff
@@ -0,0 +1,22 @@
...
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    component: search
+data:
+  ELASTICSEARCH_INDEXNAME: "{{ .Values.indexName }}"
+  SPRING_ELASTICSEARCH_REST_URIS: http://{{ .Values.elasticsearch.host }}:{{ .Values.elasticsearch.port }}
+  SPRING_ACTIVEMQ_BROKERURL: nio://{{ .Release.Name }}-activemq-broker:61616
+  ALFRESCO_SHAREDFILESTORE_BASEURL: http://{{ template "alfresco.shortname" . }}-filestore:80/alfresco/api/-default-/private/sfs/versions/1/file/
+  ALFRESCO_ACCEPTEDCONTENTMEDIATYPESCACHE_BASEURL: http://{{ template "content-services.shortname" . }}-transform-misc/transform/config
+  ALFRESCO_REINDEX_PATHINDEXINGENABLED: {{ .Values.reindexing.pathIndexingEnabled | quote }}
+  SPRING_DATASOURCE_URL: jdbc:postgresql://{{ .Release.Name }}-{{ .Values.reindexing.postgresql.hostname }}:{{ .Values.reindexing.postgresql.port | default 5432 }}/{{ .Values.reindexing.postgresql.database }}
```
alxgomz, line 17: won't work for external databases: 
https://github.com/Alfresco/acs-deployment/blob/a71275ff3782ff3f5cf6119c8ced282abb377a3c/helm/alfresco-content-services/templates/config-repository.yaml#L131

alxgomz (changes requested):
one last change which I don't think has any side effect

.pre-commit-config.yaml
```diff
@@ -25,14 +24,9 @@ repos:
     rev: v2.1.0
     hooks:
       - id: docker-compose-check
-  - repo: https://github.com/sirosen/check-jsonschema
-    rev: 0.14.2
+  - repo: https://github.com/bridgecrewio/checkov.git
```
mteodori, line 27: better move the changes related to checkov to a separate PR fixing OPSEXP-1387

helm/alfresco-content-services/Chart.yaml
```diff
@@ -14,6 +14,8 @@ home: https://www.alfresco.com
 sources:
   - https://github.com/Alfresco/acs-deployment
 dependencies:
+  - name: alfresco-common
```
mteodori, line 17: this is better named as alfresco-content-common if this subchart is specific to content

.github/workflows/verify.yml
```diff
@@ -16,10 +16,10 @@ jobs:
   helm:
     needs: [ pre-commit ]
     strategy:
-      max-parallel: 1 #cant add more parallel, problem with loadbalancers
+      max-parallel: 1 # otherwise flaky problems with aws ELB
       fail-fast: false
       matrix:
-        acs_version: [ 6.2.N, 7.0.N, 7.1.N, community, latest ]
+        acs_version: [ latest, 7.1.N, 7.0.N, 6.2.N, community ]
```
mteodori, line 22: what is the reason for changing the order here?

helm/alfresco-content-services/charts/alfresco-elasticsearch-connector/values.yaml
```diff
@@ -0,0 +1,57 @@
+image:
+  repositoryPrefix: quay.io/alfresco/alfresco-elasticsearch
+  pullPolicy: IfNotPresent
+  tag: "3.0.0"
```
mteodori, line 4: no quotes to be consistent with other tags

helm/alfresco-content-services/charts/alfresco-elasticsearch-connector/values.yaml
```diff
@@ -0,0 +1,57 @@
+image:
+  repositoryPrefix: quay.io/alfresco/alfresco-elasticsearch
```
mteodori, line 2: please use explicit repository and tag variables per each image in order to keep FluxCD happy without extra configuration

mteodori (changes requested):

helm/alfresco-content-services/Chart.yaml
```diff
@@ -34,6 +36,14 @@ dependencies:
...
     condition: alfresco-search.enabled
+  - name: alfresco-elasticsearch-connector
+    version: 0.1.0
+    condition: alfresco-elasticsearch-connector.enabled
   - name: alfresco-sync-service
     version: 3.0.9
+  - name: elasticsearch
+    alias: elasticsearch
+    repository: https://helm.elastic.co
+    version: 7.10.1
```
gionn, line 47: @jesty check, is there a newer tested version?

PR #695, opened 2022-04-21
OPSEXP-1221: Fix publish on master branch

Mieszko96:


.github/workflows/verify.yml
```diff
@@ -104,7 +104,7 @@ jobs:
         run: |
           MSG=$(git log --format=%B -n 1 ${{github.event.after}})
           echo COMMIT_MESSAGE=\"$MSG\" >> $GITHUB_ENV
-      - name: test
+      - name: Publish
         run: .github/scripts/publish_chart.sh
         shell: bash
-        if: github.ref == 'refs/heads/main' || env.COMMIT_MESSAGE == '[publish]'
+        if: github.ref == 'refs/heads/${{ github.event.repository.default_branch }}' || env.COMMIT_MESSAGE == '[publish]'
```
gionn, line 110: you can try to move the first part of the `if` at the job level to avoid at least spinning builds during PRs?

gionn: sorry, didn't see the OR and the OR looks like a feature: 

https://github.com/Alfresco/acs-deployment/commit/fb8212d83ab985364cdb9dc4eccef908596e8ba1

in any case, you can make it shorter with `github.ref_name`

is not in many examples because it's recent see [docs](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context)

alxgomz (approved):
LGTM

.github/workflows/verify.yml
```diff
@@ -104,7 +104,7 @@ jobs:
         run: |
           MSG=$(git log --format=%B -n 1 ${{github.event.after}})
           echo COMMIT_MESSAGE=\"$MSG\" >> $GITHUB_ENV
-      - name: test
+      - name: Publish
         run: .github/scripts/publish_chart.sh
         shell: bash
-        if: github.ref == 'refs/heads/main' || env.COMMIT_MESSAGE == '[publish]'
+        if: github.ref_name == '${{ github.event.repository.default_branch }}' || env.COMMIT_MESSAGE == '[publish]'
```
gionn, line 110:
```suggestion
        if: github.ref_name == github.event.repository.default_branch || env.COMMIT_MESSAGE == '[publish]'
```

it's not very intuitive, but `if` are implicitly an expression and not a string, see https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif

gionn (approved):

mteodori (approved):

PR #705, opened 2022-05-23
OPSEXP-936 Switch to statefulset for search

gionn:
OPSEXP-936


helm/alfresco-content-services/charts/alfresco-search/templates/statefulset.yaml
```diff
@@ -10,15 +10,12 @@ metadata:
     heritage: {{ .Release.Service }}
     component: search
 spec:
-  replicas: {{ .Values.replicaCount }}
+  replicas: 1
```
gionn, line 13: this is not really supported since we have a single and static pvc (every pod get the same volume and this is not how it's supposed to work)

PR #707, opened 2022-05-24
Use built helm-docs

gionn:


.pre-commit-config.yaml
```diff
@@ -2,7 +2,7 @@ repos:
   - repo: https://github.com/norwoodj/helm-docs
     rev: v1.8.1
     hooks:
-      - id: helm-docs
+      - id: helm-docs-built
```
gionn, line 5: requires a `brew install go` on the host instead of helm-docs itself but at least the helm-docs version is the one specified here

helm/alfresco-content-services/README.md
```diff
@@ -1,6 +1,4 @@
 
-----------------------------------------------
-Autogenerated from chart metadata using [helm-docs v1.8.1](https://github.com/norwoodj/helm-docs/releases/v1.8.1)
```
gionn, line 3: probably a side effect of building from source instead of using a release :\

PR #714, opened 2022-06-14
OPSEXP-1523 Fix chart publish

gionn:
OPSEXP-1523

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -20,8 +20,8 @@ jobs:
     steps:
       - id: branch
         run: |
-          BRANCH_REF=${GITHUB_HEAD_REF:-$GITHUB_REF}
-          BRANCH_NAME=${BRANCH_REF##*\/}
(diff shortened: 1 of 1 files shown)


.github/workflows/verify.yml
```diff
@@ -16,12 +16,13 @@ jobs:
...
     outputs:
-      branch_name: ${{ steps.branch.outputs.branch_name }}"
+      branch_name: ${{ steps.branch.outputs.branch_name }}
     steps:
       - id: branch
+        name: Evaluate branch name and set as output
         run: |
-          BRANCH_REF=${GITHUB_HEAD_REF:-$GITHUB_REF}
-          BRANCH_NAME=${BRANCH_REF##*\/}
+          BRANCH_NAME=${GITHUB_HEAD_REF:-$GITHUB_REF_NAME}
```
mteodori, line 24: I don't think this is correct, as _REF have refs/heads/ before while NAME only has the branch name, sadly there is no GITHUB_HEAD_REF_NAME var yet

mteodori (approved):

PR #715, opened 2022-06-14
OPSEXP-1494 Release acs-deployment 5.3.0-M2

mteodori:
Ref. OPSEXP-1494

waiting for updates of all of https://alfresco.atlassian.net/wiki/spaces/TECH/pages/932218719/Releases+in+Tilehurst


alxgomz (changes requested):
Shouldn't we wait for admin-app 7.4 and adw 2.9.0 and release every new component?
Also as ACS 7.3 introduce support for postgresql 14.3 try and bump the postgresql image version to that

mteodori:
> Shouldn't we wait for admin-app 7.4 and adw 2.9.0 and release every new component?

yes, I put in description but better set as draft to avoid confusion

> Also as ACS 7.3 introduce support for postgresql 14.3 try and bump the postgresql image version to that

no need for that if the former version is supported but if it helps with the chart maintenance we can do

alxgomz:
> > Shouldn't we wait for admin-app 7.4 and adw 2.9.0 and release every new component?
> 
> yes, I put in description but better set as draft to avoid confusion
> 
> > Also as ACS 7.3 introduce support for postgresql 14.3 try and bump the postgresql image version to that
> 
> no need for that if the former version is supported but if it helps with the chart maintenance we can do

my point was more about aligning charts to the supported stack so only using 14.3 image for postgres. Updating the chart should be part of a different task as it'll require modification to the values file too

montgolfiere (changes requested):
As per end of last week, we now also have FE releases such as:

- alfresco-digital-workspace:2.9.0
- alfresco-admin-app:7.4.0
- alfresco-content-app:2.11.0 (eg. at least for community-docker-compose.yml)

alxgomz (changes requested):
Can you please update the main READ file (table in the "Versioning" section)

mteodori:
> Can you please update the main READ file (table in the "Versioning" section)

isn't already updated?

PR #718, opened 2022-06-16
717 _helpers-ingress.tpl: use - indicator with Literal Block Scalar

### LINKED ISSUE #717: _helpers-ingress.tpl: use - indicator with Literal Block Scalar
While customizing annotations for ingress in the repository:
```
repository:
  ingress:
    annotations:
      cert-manager.io/cluster-issuer: "letsencrypt"
```
the CertificateRequest complained that it cannot find cluster issuer `"letsencrypt\n"`

**Solution?**
I think the **Literal Block Scalar** should be replaced with **– indicator with Literal Block Scalar** in ``_helpers-ingress.tpl``
```
  {{- $annotation | nindent 4 }}: |-
    {{- $value | nindent 6 }}
```
This way, the newline at the end of the string is removed (while keeping the newlines between the lines)

dboldureanu:
Fixes #717


gionn:
Closing since the linked issue has been resolved with another PR.

PR #727, opened 2022-07-04
OPSEXP-799 Make repository and filestore claims configurable

### LINKED ISSUE #203: volume claim alfresco-volume-claim assigned to multiple pods
Almost all pods (except solr but including e.g. activemq) are using the same volume claim. This is not possible in the infrastructure of the Alfresco customer I am working with which is based on vmware vsphere. At the moment only the postgresql volume claim can be changed by setting the property postgresql.persistence.existingClaim. All other pods are sharing the same configuration property persistence.existingClaim which makes it impossible to configure different claims without modifying the chart templates.

We would like to have an option to configure different claims for all pods that does not require forking and modifying the helm charts.

gionn:
Fix #203 
OPSEXP-799


alxgomz (commented):
I really don't like the fact we have this `persistence` root value. It works the exact opposite way subcharts work.
I would align behavior in order to fix that. Root `persistence` would only make sense as a shared volume across services but we know that doesn't work and nobody really want that.

gionn:
I don't like it either but I was scared about breaking retrocompatibility, if users has changed the value of `persistence.existingClaim` they will need to set those new values otherwise their installation is going to break.

I can't find any helm way to mark a value as deprecated so clearly stating it in the docs is the only viable option.

gionn:
we are going to implement a consistent approach to persistent storage in our charts (OPSEXP-1582) in the near future so we prefer to avoid providing here a partial solution for that, and that has some retro-compatibility issues we prefer to tackle later

PR #731, opened 2022-07-07
OPSEXP-1574: allow specifying an external ActiveMQ in search enterprise

alxgomz:
Ref: OPSEXP-1574


helm/alfresco-content-services/values.yaml
```diff
@@ -598,6 +598,7 @@ alfresco-search:
 
 alfresco-elasticsearch-connector:
   enabled: false
+  messageBroker: *acs_messageBroker
```
gionn, line 601: how is this going to work when using internal activemq?

alxgomz: in-cluster activemq creation rely on `activemq.enabled` and the default values will provide an object with null properties so `if/else` flow control and `default` function should kickin appropriately.

gionn: for all the templates in the main chart it's fine, but I was wondering here in the es subchart how it can work, since you are explicitly referencing the messageBroker object that will be empty when using in-cluster activemq (so null values are passed to the subchart and the subchart doesn't know anything about activemq.enabled)

alxgomz: es subcharts - and all subcharts in general - should depend on alfresco-content-common which provide a common named templates to provide pods with appropriate env vars.
This named template leverages a pipeline in containing the `default` function. So in case `messageBroker.url` is null it'll use the expected name of the in-cluster activemq service: https://github.com/Alfresco/acs-deployment/blob/054704fca2e735447b9878f7c503a0639019ce52/helm/alfresco-content-services/charts/alfresco-content-common/templates/_helpers-activemq.tpl#L15

gionn (approved):

PR #743, opened 2022-07-27
OPSEXP-1587: Sync service to use external URL for ActiveMQ

alxgomz:
Ref: OPSEXP-1587

### DIFF
--- a/docs/helm/examples/with-aws-services.md
+++ b/docs/helm/examples/with-aws-services.md
@@ -111,7 +111,7 @@ In order to use the S3 connector and external database options, the S3 connector
 To use the S3 connector, RDS and Amazon MQ we have to disable the internal default components via the Helm "set" command and additionally provide the service endpoints and credentials we made a note of in the previous sections.
 
 When we bring all this together we can deploy ACS using the command below (replacing all the `YOUR-XZY` properties with the values gathered during the setup of the services).
-Edit your `values.yml`file so it contains below elements:
+Edit your `values.yml` file so it contains below elements:
 
 ```yaml
 externalPort: 443
@@ -152,11 +152,10 @@ Then you can deploy using:
 helm -n alfresco install acs ./alfresco/alfresco-content-services --atomic --timeout 10m0s
 ```
 
-If you're deploying from the registry of charts you can't update the `values.yml`file. Instead you either:
+If you're deploying from the registry of charts you can't update the `values.yml` file. Instead you either:
 
-* use a local copy of the `values.yml`file amended as shown above
-
-* use `--set` options to pass individual values.
+    * use a local copy of the `values.yml` file amended as shown above
+    * use `--set` options to pass individual values.
 
 Note however that the main `values.yml` file uses [YAML anchors and aliases](https://yaml.org/spec/1.2.2/#3222-anchors-and-aliases) to propagate configuration to subcharts if needed.
 Using `--set` do not fill the yaml file with passed values and so aliases will never point to a node filled with the values passed by the command line.
--- a/docs/helm/examples/with-aws-services.md
+++ b/docs/helm/examples/with-aws-services.md
@@ -86,7 +86,7 @@ The following sections describe how to setup the AWS services and highlights the
 
 1. Create an Amazon MQ broker using the "Create brokers" wizard in the [MQ Console](https://console.aws.amazon.com/amazon-mq/home).
 
-    * Select "Single-instance broker" option and press the Next button
+    * Select "Single-instance broker" or "Active/standby broker" option and press the Next button
     * Provide a "Broker name" of your choosing
     * In the "ActiveMQ Access" section specify `alfresco` as the "Username" and a "Password" of your choice
     * In the "Additional settings" section choose the "Select existing VPC and subnet(s)" option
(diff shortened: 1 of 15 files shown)


docs/helm/examples/with-aws-services.md
```diff
@@ -102,16 +102,67 @@ The following sections describe how to setup the AWS services and highlights the
...
+  * use a local copy of the `values.yml` file amended as shown above
+  * use `--set` options to pass individual values.
+
+Note however that the main `values.yml` file uses [YAML anchors and aliases](https://yaml.org/spec/1.2.2/#3222-anchors-and-aliases) to propagate configuration to subcharts if needed.
+Using `--set` do not fill the yaml file with passed values and so aliases will never point to a node filled with the values passed by the command line.
+That means when using command line you must also pass aliased values as shown below (at the end of the command):
 
 ```bash
-helm install acs alfresco/alfresco-content-services \
+helm -n alfresco install acs alfresco-stable/alfresco-content-services --atomic --timeout 10m0s
```
gionn, line 165: alfresco-stable?

docs/helm/examples/with-aws-services.md
```diff
@@ -102,16 +102,67 @@ The following sections describe how to setup the AWS services and highlights the
...
+  password: YOUR-MQ-PASSWORD
+```
+
+Then you can deploy using:
+
+```bash
+helm -n alfresco install acs ./alfresco/alfresco-content-services --atomic --timeout 10m0s
+```
+
+If you're deploying from the registry of charts you can't update the `values.yml` file. Instead you either:
```
gionn, line 155: looks like there is `-f` argument for helm install for providing a path to a values file

helm/alfresco-content-services/charts/alfresco-sync-service/templates/config-syncservice.yaml
```diff
@@ -10,6 +10,7 @@ metadata:
     chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
     component: syncservice
 data:
+  {{ template "spring.activemq.config" . }}
```
gionn, line 13: is syncservice a spring boot application? (I think not)

I don't think it requires both env vars and individual properties defined at L26

helm/alfresco-content-services/charts/alfresco-sync-service/templates/deployment-syncservice.yaml
```diff
@@ -33,19 +33,34 @@ spec:
...
+            while [ -z $MQREADY ]; do
+              for broker in "${array[@]}"; do broker_socket=${broker%%\?*};
+                [[ "${broker_socket#*://}" == *":"* ]] || NCPORT=61616;
+                echo "Trying ${broker_socket#*://} $NCPORT";
+                nc -w1 ${broker_socket#*://} $NCPORT;
+                if [ $? -eq 0 ]; then echo "ok";
+                  MQREADY=1;
+                  break;
+                else echo "not ok";
+                  sleep 5;
```
gionn, line 60: if any of the broker(s) is reachable then it is ok?

I was thinking to wait for all the brokers to be available, but maybe it's fine as it is 👀

helm/alfresco-content-services/charts/alfresco-sync-service/values.yaml
```diff
@@ -116,8 +115,8 @@ postgresql-syncservice:
 initContainers:
   activemq:
     image:
-      repository: busybox
-      tag: 1.35.0
+      repository: bash
+      tag: 5.1.16
```
gionn, line 119: welcome to the future!

docs/helm/examples/with-aws-services.md
alxgomz: Yes but unfortunatelly anchors and aliases only live within a single file

helm/alfresco-content-services/charts/alfresco-sync-service/templates/config-syncservice.yaml
alxgomz: No, it's not. I'm just reusing them to pass the value to the command line argument... Though I'm not sure why I didn't use the ACTIVEMQ_* ones instead.. 🤔. Will try that as it would be less confusing

alxgomz: Ah I know remember why I did use the SPRING vars:
the activemq.config named template has a condition which rely on an activemq.enable value node. I kept this to avoid introducing more changes in the main chart but this node is not available to subcharts and I didn't want to have another template that would essentially do the same thing as spring.activemq.config just with different variable names.
I also don't want to spread activemq.enable to subcharts as 1) it doesn'treally make sense for them in their own context. é) I think we should skip a to model where feature flip relies on configuration being null or not (e.g. here messageBroker.url: null == activemq.enable: true).

helm/alfresco-content-services/charts/alfresco-sync-service/templates/deployment-syncservice.yaml
alxgomz: Yes AWS provides an hot-standby approach for ActiveMQ brokers (among other setups), where only one host is reachable at a time.

docs/helm/examples/with-aws-services.md
alxgomz: that used to be how the doc referenced the alfresco helm repo but looks like it's been changed to `alfresco` when we decided to give up on incubator repo.
Changing that.

helm/alfresco-content-services/charts/alfresco-sync-service/templates/config-syncservice.yaml
gionn: ok it's fine, just a little confusing to see the `spring.activemq.config` template used there.

Maybe an idea could be to define another template very similar to `spring.activemq.config` that provides env vars with the same values but with different name so it doesn't really look like we are using spring boot here

but since it's just for this maybe it's not worth the hassle

alxgomz: I decided it was not worth it indeed

gionn: > I didn't want to have another template that would essentially do the same thing as spring.activemq.config just with different variable names.

ok you already think about that xD

gionn (approved):

PR #744, opened 2022-08-01
OPSEXP-1627 Add support for Elasticsearch credentials

gionn:
OPSEXP-1627


helm/alfresco-content-services/values.yaml
```diff
@@ -890,3 +896,16 @@ global:
 #   region:
 #   s3Bucket:
 #   comprehendRoleARN:
+
+  # -- Shared connections details for Elasticsearch/Opensearch, required when alfresco-elasticsearch-connector.enabled is true
+  elasticsearch:
```
Mieszko96, line 901: It's really not good that we have elasticseatch values in 3 places. 901line 736 and 88, but don't know how to do it better :(

Mieszko96 (approved):

helm/alfresco-content-services/values.yaml
gionn: 736 refers to the bitnami elasticsearch chart configuration, while the others are for how helm globals works (they need to be declared in each chart)

helm/alfresco-content-services/charts/alfresco-elasticsearch-connector/values.yaml
```diff
@@ -80,3 +84,15 @@ messageBroker:
...
+    # -- The host where service is available
+    host: elasticsearch-master
+    # -- The port where service is available
+    port: 9200
+    # -- Valid values are http or https
+    protocol: http
+    # -- The username required to access the service, if any
+    user: null
+    # -- The password required to access the service, if any
+    password: null
```
alxgomz, line 98: To me these should not be part of `values` and should instead be expanded from a named template that's part of `alfresco-content-common` in case no other detail is provided by users

helm/alfresco-content-services/templates/config-repository.yaml
```diff
@@ -51,7 +51,10 @@ data:
       {{- end }}
       {{- if index .Values "alfresco-elasticsearch-connector" "enabled" }}
       -Delasticsearch.createIndexIfNotExists=true
-      -Delasticsearch.host={{ index .Values "alfresco-elasticsearch-connector" "elasticsearch" "host" }}
+      -Delasticsearch.host={{ index .Values "alfresco-elasticsearch-connector" "elasticsearch" "host" | default .Values.global.elasticsearch.host }}
+      -Delasticsearch.port={{ index .Values "alfresco-elasticsearch-connector" "elasticsearch" "port" | default .Values.global.elasticsearch.port }}
+      -Delasticsearch.user={{ index .Values "alfresco-elasticsearch-connector" "elasticsearch" "user" | default .Values.global.elasticsearch.user }}
+      -Delasticsearch.password={{ index .Values "alfresco-elasticsearch-connector" "elasticsearch" "password" | default .Values.global.elasticsearch.password }}
```
alxgomz, line 57: I'd rather have that part of a named tpl as well

helm/alfresco-content-services/values.yaml
```diff
@@ -890,3 +896,16 @@ global:
...
+    # -- The host where service is available. The provided default is for when elasticsearch.enabled is true
+    host: elasticsearch-master
+    # -- The port where service is available
+    port: 9200
+    # -- Valid values are http or https
+    protocol: http
+    # -- The username required to access the service, if any
+    user: null
+    # -- The password required to access the service, if any
+    password: null
```
alxgomz, line 911: same here. I would have only null values (so you can document values) and use a template that basically outputs:
`es.*` if exists otherwise `global.es.*` otherwise default values as they are computed in existing tpl

helm/alfresco-content-services/templates/config-repository.yaml
gionn: even if we are using it only in one place?

alxgomz: not as useful of course but I would not consider it's a problem. It would provide a centralised place where we compute ES details.
Can also be done as part of another ticket up to you

alxgomz (commented):
I don't want to bother more on this task but I realize we have credentials set in cm instead of secrets. Can you please create another ticket to move creds to secrets where needed (e.g. activemq)

gionn:
> I don't want to bother more on this task but I realize we have credentials set in cm instead of secrets. Can you please create another ticket to move creds to secrets where needed (e.g. activemq)

https://alfresco.atlassian.net/browse/OPSEXP-1651

helm/alfresco-content-services/templates/config-repository.yaml
gionn: I think there is a problem with the approach of moving to a single tpl in the alfresco-content-common subchart: there are some values that make sense only in the main chart and that never exists in other subcharts or they are somewhere else. For `_helpers-activemq.tpl` worked because you are in the same chart so Values are consistent. With elasticsearch instead we have different charts and there is also a clash in Values: subchart have .Values.elasticsearch to provide overrides while the main chart at the same path has the section to configure the elasticsearch bitnami chart.

I think it's better to better discuss approach for the future, for now I suggest to move as is if we want this part of the next release.

gionn: In the end I've created a new tpl but that lives in the parent chart so it doesn't imply can be used in every chart the depends on alfresco-content-common

mteodori (approved):

PR #747, opened 2022-08-02
OPSEXP-1546 Try KinD

gionn:
OPSEXP-1546


gionn:
kind support for community is now on master

PR #748, opened 2022-08-04
Fix imagePullSecret handling when secret is not set

dpmerron-ltd:
When performing a helm upgrade over an existing deployment helm would error with; 

client.go:205: [debug] error updating the resource "alfresco-content-services-alfresco-router":
	 failed to create patch: map: map[] does not contain declared merge key: name
upgrade.go:355: [debug] warning: Upgrade "alfresco-content-services" failed: failed to create patch: map: map[] does not contain declared merge key: name

This is because the original deployment would set the imagePullSecret to "name: " and would try to be re-set at upgrade. Since a blank value is defaulted in the chart values this conditional is needed in all the deployment (and is present in the others).

### DIFF
--- a/helm/alfresco-content-services/charts/alfresco-sync-service/templates/deployment-syncservice.yaml
+++ b/helm/alfresco-content-services/charts/alfresco-sync-service/templates/deployment-syncservice.yaml
@@ -30,8 +30,11 @@ spec:
     {{- if .Values.nodeSelector }}
       nodeSelector: {{- .Values.nodeSelector | toYaml | nindent 8 }}
     {{- end }}
+      {{- if .Values.global.alfrescoRegistryPullSecrets }}
+      # only set this secret if a private docker registry variable is defined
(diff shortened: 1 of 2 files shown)


gionn:
Hello, would you mind fixing also `helm/alfresco-content-services/charts/alfresco-sync-service/templates/deployment-syncservice.yaml` with the same approach?

for the future we really need to define a new template with that snippet and reuse everywhere

gionn:
Looks like we have some issues on github actions for PR coming from forked repos, in the meanwhile I've manually run your code [here](https://github.com/Alfresco/acs-deployment/actions/runs/2797788975) so it's fine to merge.

@dpmerron-ltd please [sign the CLA](https://github.com/Alfresco/acs-deployment/pull/748#issuecomment-1205021387) and we can proceed to merge

gionn (approved):

gionn:
Thanks for contributing! 🌮

PR #750, opened 2022-08-08
OPSEXP-1670 Support existing secret for message broker credentials

barrymars:
instead of manualling setting credentials in values file


barrymars:
https://github.com/Alfresco/collaborating-with-alfresco/issues/4

gionn:
Hello @barrymars and thanks for raising this PR.

We are aware that those charts lacks a proper support of kubernetes secrets as a common standard way to provide sensitive values and we have this kind of improvements on our roadmap (OPSEXP-1651)

The proper approach should be to have a kubernetes secret template populated with the values we currently have (messageBroker.user and messageBroker.password) OR provide a kubernetes secret name like you did.

In this way every template would always reference a kubernetes secret, either managed by the chart itself or an externally managed like you need.

With this requisites in mind, I think we can't yet accept the PR as it is now but I am open to listen for others opinion on that (cc @alxgomz)

barrymars:
@gionn ok, I can look at expanding my changes to include that, but will take a bit longer

had hoped to add a little bit of functionality without completely rewriting how secrets are handled :)

barrymars:
ok, so didn't take as long as i thought

haven't had a chance to test this yet, but was this more what you were thinking @gionn ?

gionn:
> haven't had a chance to test this yet, but was this more what you were thinking @gionn ?

indeed it was! I made a quick review and looks fine, now we should proceed to get a green build.

helm/alfresco-content-services/charts/alfresco-content-common/templates/_helpers-activemq.tpl
```diff
@@ -7,24 +7,13 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
...
-  ACTIVEMQ_USER: {{ .Values.activemq.adminUser.username | quote | default "" }}
-  ACTIVEMQ_PASSWORD: {{ .Values.activemq.adminUser.password |  quote | default "" }}
-{{- else -}}
-  ACTIVEMQ_URL: {{ .Values.messageBroker.url | quote }}
-  ACTIVEMQ_USER: {{ .Values.messageBroker.user | quote | default ""  }}
-  ACTIVEMQ_PASSWORD: {{ .Values.messageBroker.password | quote | default "" }}
-{{- end -}}
+  ACTIVEMQ_URL: "$BROKER_URL"
+  ACTIVEMQ_USER: "$BROKER_USERNAME"
+  ACTIVEMQ_PASSWORD: "$BROKER_PASSWORD"
```
gionn, line 12: I've done some manual testing and this is not supported by kubernetes as it is.

Looking at https://kubernetes.io/docs/tasks/inject-data-application/define-interdependent-environment-variables/#define-an-environment-dependent-variable-for-a-container it looks like that to properly reference another variabile value is to define the helper block like:

```helm
{{- define "activemq.config" -}}
- name: ACTIVEMQ_URL
  value: $(BROKER_URL)
- name: ACTIVEMQ_USER
  value: $(BROKER_USERNAME)
- name: ACTIVEMQ_PASSWORD
  value: $(BROKER_PASSWORD)
{{- end -}}
```

and then reference that within `spec.containers.env` of deployments like:

```helm
          envFrom:
          - configMapRef:
              # config map to use, defined in config-transform-router.yaml
              name: {{ template "alfresco.shortname" . }}-router-configmap
          - secretRef:
              name: {{ default (printf "%s-brokersecret" (include "content-services.shortname" .)) .Values.messageBroker.existingSecretName }}
          env:
            {{- include "activemq.config" . | nindent 12 }}
```

and of course remove `activemq.config` from the configmap template that become unnecessary.

Would you mind trying with this approach?

gionn (changes requested):

helm/alfresco-content-services/charts/alfresco-content-common/templates/_helpers-activemq.tpl
barrymars: @gionn sorry about that, should have remembered that env var trick doesn't work via configmaps

was thinking of making similar changes for the other secrets in the chart....would you prefer those in their own MRs or lumped within this one?

gionn: in the long run we want to use this approach for every secrets, but for this PR just focus on broker credentials so we can merge earlier

barrymars: ok, once this is approved i'll create separate MRs for the other ones I come across as we roll it out for our purposes, thanks

barrymars: @gionn saw the pipeline was failing on the `sync-service` not being able to connect to ActiveMQ

pretty sure this is because the `sync-service` configmap uses the secret env variables to pass into `JAVA_OPTS` and so isn't expanding them...got a thought about how to fix this that i will push momentarilly

will also rename "activemq.config" to "activemq.env" to make it clearer what it's used for

gionn:
I've pushed two more commits, now I have everything green on a local deployment, now the plan is:

- [ ] green CI
- [ ] another round of review
- [ ] testing on our internal pipeline

helm/alfresco-content-services/charts/alfresco-elasticsearch-connector/templates/reindexing-job.yaml
```diff
@@ -42,6 +42,7 @@ spec:
                 secretKeyRef:
```
gionn, line 42: secretRef is missing

gionn (changes requested):

helm/alfresco-content-services/charts/alfresco-content-common/templates/_helpers-activemq.tpl
```diff
@@ -6,25 +6,20 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
...
-  SPRING_ACTIVEMQ_PASSWORD: {{ .Values.messageBroker.password | quote }}
-  {{- end }}
+{{- define "spring.activemq.env" -}}
+- name: JAVA_OPTS
+  value: >-
+    $(JAVA_OPTS)
+    -Dmessaging.broker.url=$(BROKER_URL)
+    -Dmessaging.username=$(BROKER_USERNAME)
+    -Dmessaging.password=$(BROKER_PASSWORD)
 {{- end -}}
```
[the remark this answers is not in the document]
barrymars: are there environment variables that can be directly consumed by spring boot, instead of having to set `-Dmessaging.broker.url` etc in JAVA_OPTS ?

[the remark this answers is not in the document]
gionn: the ones that you see where in use previously: SPRING_ACTIVEMQ_BROKERURL, SPRING_ACTIVEMQ_USER, SPRING_ACTIVEMQ_PASSWORD.

barrymars:
@gionn sorry it's taken a while to get back to this....combination of work pressures and PTO

replied and fixed those last few issues you found

gionn:
no worries, same situation here 😀 
I will take a look hopefully later today or start of the next week and then we should be able to merge 🚀

gionn (approved):

alxgomz (approved):
Good stuff! Thanks for your contribution.
Just have one concern we'll address later. Tracked in https://alfresco.atlassian.net/browse/OPSEXP-1729

PR #751, opened 2022-08-09
OPSEXP-1665 Fixup default values when using embedded elasticsearch

gionn:
OPSEXP-1665

We can have computed values when generating templates in the main chart but not for values that needs to be provided to a subchart. I revert to having es connection defaults for embedded elasticsearch that is the default.


kavitshah-gl (approved):
Looks fine.

kklassa (approved):

PR #755, opened 2022-08-16
OPSEXP-1656 avoid duplication when declaring imagePullSecrets in helm charts

batonck:
OPSEXP-1656


.github/workflows/verify.yml
```diff
@@ -5,6 +5,7 @@ on:
       - 'master'
   push:
     branches:
+      - 'OPSEXP-1656-helm-charts'
```
gionn, line 8:
```suggestion
```

no need for this when you also open a PR (marked as draft initially)

helm/alfresco-content-services/charts/alfresco-content-common/templates/_helpers-image-pull-secrets.tpl
```diff
@@ -0,0 +1,7 @@
+{{- define "alfresco-content-services.imagePullSecrets" }}
+{{- if .Values.global.alfrescoRegistryPullSecrets }}
+# only set this secret if a private docker registry variable is defined
+imagePullSecrets:
+    - name: {{ .Values.global.alfrescoRegistryPullSecrets }}
+{{- end }}
+{{- end }}
```
gionn, line 7: alfresco-cotent-common subchart should already be dependency of all the other subcharts, so in the end it's only needed to be defined once (and we don't want to duplicate the same things in so many files anway)

gionn (changes requested):
please set a proper PR title describing the enhancement it provides (usually the jira title is fine)

gionn (commented):
see failure in https://github.com/Alfresco/acs-deployment/runs/7878016685?check_suite_focus=true#step:8:89

helm/alfresco-content-services/charts/alfresco-search/templates/_helpers-image-pull-secrets.tpl
```diff
@@ -0,0 +1,7 @@
+{{- define "alfresco-content-services.imagePullSecrets" }}
+{{- if .Values.global.alfrescoRegistryPullSecrets }}
+# only set this secret if a private docker registry variable is defined 2
+imagePullSecrets:
+  - name: {{ .Values.global.alfrescoRegistryPullSecrets }}
+{{- end }}
+{{- end }}
```
gionn, line 7: to get rid of this file I think you just need to add in the alfresco-search [Chart.yml](https://github.com/Alfresco/acs-deployment/blob/master/helm/alfresco-content-services/charts/alfresco-search/Chart.yaml) the dependency to alfresco-content-common like has been done in another subchart like [here](https://github.com/Alfresco/acs-deployment/blob/master/helm/alfresco-content-services/charts/alfresco-elasticsearch-connector/Chart.yaml#L8)

gionn (changes requested):

helm/alfresco-content-services/charts/alfresco-content-common/templates/_helpers-image-pull-secrets.tpl
```diff
@@ -0,0 +1,7 @@
+{{- define "alfresco-content-services.imagePullSecrets" }}
+{{- if .Values.global.alfrescoRegistryPullSecrets }}
+# only set this secret if a private docker registry variable is defined 1
```
gionn, line 3:
```suggestion
```

not sure what was the reason for that but don't seems very useful and worth keeping

gionn (changes requested):
one last small thing and for me it's fine

gionn (approved):

PR #759, opened 2022-08-25
OPSEXP-1688: Update docker compose to ACS 7.2.1 and 7.3.0-M2

piraartur:
OPSEXP-1688: Updating docker compose files and markdown table in readme.md

### DIFF
--- a/docker-compose/7.1.N-docker-compose.yml
+++ b/docker-compose/7.1.N-docker-compose.yml
@@ -12,7 +12,7 @@ version: "2"
 
 services:
   alfresco:
-    image: quay.io/alfresco/alfresco-content-repository:7.1.1.5
+    image: quay.io/alfresco/alfresco-content-repository:7.1.1.3
     mem_limit: 1900m
     environment:
       JAVA_TOOL_OPTIONS: "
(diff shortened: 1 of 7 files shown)


docker-compose/docker-compose.yml
```diff
@@ -12,7 +12,7 @@ version: "2"
 
 services:
   alfresco:
-    image: quay.io/alfresco/alfresco-content-repository:7.3.0-M1
+    image: quay.io/alfresco/alfresco-content-repository:7.3.0-A12
```
gionn, line 15: stick to -M versions, -A are alphas

docker-compose/7.1.N-docker-compose.yml
```diff
@@ -12,7 +12,7 @@ version: "2"
 
 services:
   alfresco:
-    image: quay.io/alfresco/alfresco-content-repository:7.1.1.3
+    image: quay.io/alfresco/alfresco-content-repository:7.1.2-8269
```
gionn, line 15: not sure which is the versioning scheme with the old 7.1 but I would say `-dddd` are dev versions, so latest available should be:

```suggestion
    image: quay.io/alfresco/alfresco-content-repository:7.1.1.5
```

gionn (changes requested):

docker-compose/docker-compose.yml
alxgomz: also `share` needs to be updated to match repo image

alxgomz (changes requested):

gionn (approved):

alxgomz (approved):
LGTM

PR #762, opened 2022-09-01
OPSEXP-1575: move PGDATA away from efs default volume claim

alxgomz:
Ref: OPSEXP-1575


kklassa (approved):
to my knowledge, looks good

helm/alfresco-content-services/values.yaml
```diff
@@ -687,7 +687,12 @@ postgresql:
   primary:
     nodeSelector: {}
   persistence:
-    existingClaim: "alfresco-volume-claim"
```
gionn, line 690: what happens with existing deployments? would recreate the pod without the mount?

we may want to start providing an upgrade doc page where we mention how to proceed with breaking changes while upgrading released versions.

alxgomz: production deployments should not use "in-cluster" db we provide but essentially yes, that would trigger a pods recreation with a different volume depending on the chart configuration (whether static provisioning through `existingClaim` is used, or dynamic provisioning through `storageClass`).
Let me create another PR that's taking doc aspect into account

PR #767, opened 2022-09-08
OPSEXP-1545 Decrease duplication in build scripts for helm charts with github custom actions

kklassa:
Making use of new actions created as part of OPSEXP-1545


alxgomz (approved):
LGTM

PR #769, opened 2022-09-12
OPSEXP-1717 Add Elasticsearch connector to docker-compose

gionn:
OPSEXP-1717

Test with:

```
docker-compose -f docker-compose.yml -f elasticsearch-override-docker-compose.yml up -d
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,6 +24,7 @@ repos:
     rev: v2.1.0
     hooks:
       - id: docker-compose-check
+        exclude: override-docker-compose.yml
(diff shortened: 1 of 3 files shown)


alxgomz (changes requested):
I think we're missing the reindex container (this one is not part of aio)

gionn:
> I think we're missing the reindex container (this one is not part of aio)

not sure 100% that it make sense to have it here (since we don't have persistence so we always start with a clean repo) but since it's not harming I think we can have it

alxgomz (approved):
LGTM

PR #770, opened 2022-09-15
Opsexp 1447 alfcommon refactored

alxgomz:
Ref: OPSEXP-1447

* split alfresco-common to a separate chart (holding common external dependencies)
* refactor builds to allow releasing multiple dependent charts

Hope it clarifies things a little:


.github/workflows/publish.yml
```diff
@@ -0,0 +1,61 @@
+---
+name: publish
+on:
+  workflow_run:
+    workflows:
+      - helm
+      - integration-tests
```
gionn, line 7: this will trigger one time for each

.github/workflows/compose.yml
```diff
@@ -0,0 +1,34 @@
+---
```
kklassa, line 1: I would suggest changing the name of this file as vs code thinks it is a docker-compose file and messes up linting

.github/actions/publish-charts/action.yml
```diff
@@ -0,0 +1,39 @@
+---
+name: "Publish helm chart on GitHub "
+description: "Publish new version of helm chart and update repository tags"
+inputs:
+  chart:
+    description: Name of the chart to publish
+    required: true
+    type: true
```
kklassa, line 8: I think this does not belong here 😅

.github/workflows/publish.yml
alxgomz: Yes that's expected. One publish stage for libraries charts (first), then one for for application charts (second, as they most likely depend on library)

PR #773, opened 2022-09-20
OPSEXP-1447: split alfresco-common into a library chart of its own

alxgomz:
Ref: OPSEXP-1447

This PR only takes care of creating that new charts and fixing the actual issue with conflicting dependencies.
Further chagnes are needed to adapt the release build/process


helm/alfresco-content-services/Chart.yaml
```diff
@@ -14,22 +16,27 @@ home: https://www.alfresco.com
 sources:
   - https://github.com/Alfresco/acs-deployment
 dependencies:
-  - name: alfresco-content-common
-    version: 0.1.0
+  - name: alfresco-common
+    version: 0.1.0-A417
```
kklassa: just a note that after merging https://github.com/Alfresco/alfresco-build-tools/pull/120 the incubator charts by default will be created with the SNAPSHOT suffix, so like `0.1.0-SNAPSHOT`

.github/workflows/verify.yml
```diff
@@ -72,9 +74,16 @@ jobs:
...
     if: github.event.head_commit.message == '[publish]'
+    needs:
+      - helm
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        chart:
+          - alfresco-common
+          - alfresco-content-service
```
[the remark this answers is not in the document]
kklassa: also, I don't know if you intended it this way but in the current state when the matrix is defined but not referenced the action will publish the same thing twice I think

kklassa (approved):

.github/workflows/verify.yml
```diff
@@ -1,23 +1,23 @@
+---
 name: verify
 on:
   pull_request:
     branches:
       - 'master'
   push:
-    branches:
-      - 'master'
-      - pr-*
```
gionn, line 9: keep master or it will trigger 2x workflow when branch is not in sync with master

.github/workflows/verify.yml
```diff
@@ -1,23 +1,23 @@
...
 jobs:
   pre-commit:
     runs-on: ubuntu-latest
     steps:
+      - uses: hmarr/debug-action@v2.0.1
```
gionn, line 12: remove before merge

.github/workflows/verify.yml
```diff
@@ -72,9 +72,16 @@ jobs:
...
+    if: contains(github.event.head_commit.message, '[publish]')
+    needs:
+      - helm
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        chart:
+          - alfresco-common
+          - alfresco-content-services
```
gionn, line 84: it is fine to do this in parallel?

gionn (changes requested):

gionn (approved):

PR #780, opened 2022-10-07
Feature/search services replication

jvanhienen:
Search services replication

By default Alfresco [already supports](https://docs.alfresco.com/insight-engine/latest/config/replication/) master-slave replication for the search services, however it wasn't available yet within the helm chart. The implementation that I propose is based on the [docker-compose example](https://github.com/Alfresco/SearchServices/tree/master/search-services/packaging/src/docker/6.x) which resides in the search-services repository. There's also a change in the alfresco-common chart because of the function that is used to define the alfresco-search host.

We are running this in our own environment and everything seems to work like it does with the docker-compose example. If there are any suggestions, feel free to submit them. And if anything needs to be changed, please let me know.

Thank you in advance.

I've read the contribution guideline and I accept the Alfresco Contribution Agreement.

### DIFF
--- a/.github/workflows/community-compose.yml
+++ b/.github/workflows/community-compose.yml
@@ -1,0 +1,31 @@
+---
+name: Docker Compose (Community)
+on:
+  pull_request:
+    branches:
+      - 'master'
+    paths:
+      - 'docker-compose/community-docker-compose.yml'
+      - 'test/postman/docker-compose/**'
+      - '.github/workflows/community-compose.yml'
+      - '.pre-commit-config.d/compose.yaml'
+  push:
+    branches:
+      - 'master'
+jobs:
+  pre_commit:
+    name: Run pre-commit compose checks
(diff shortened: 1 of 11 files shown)


shazada:
Is there something we can do to speed it up to include this?

alxgomz:
Thank you for submitting this PR (and apologize for the late answer).
We consider Search service (Solr) not a good fit for kubernetes environment. It has different drawbacks in terms of resiliency, reliability and maintenance. We recommend deploying search outside of the kubernetes cluster and pointing the repository to that external instance in production environment. 
Therefore, we want to avoid providing high availability feature out of the box as that would lead people to think it's suitable for production in kubernetes environment.
We are thinking about making search and other components separate charts, so that would be a lot easier for you to plug ACS chart to your own search chart in which you could maintain the replication if it's enough for you. That's still something in "ideation" but we hope to achieve that in a not too distant future.

Thanks again for your contribution.

PR #782, opened 2022-10-10
OPSEXP-1746: update workflow badges

alxgomz:
Ref: OPSEXP-1746

Follow up PR as I forgot to update the badges


gionn (changes requested):
can you adjust also the workflow names so they are consistent across helm/compose?

PR #783, opened 2022-10-10
OPSEXP-1766: fix ES protocol Issue

alxgomz:
Ref: OPSEXP-1766

### DIFF
--- a/helm/alfresco-content-services/templates/_helpers-elasticsearch.tpl
+++ b/helm/alfresco-content-services/templates/_helpers-elasticsearch.tpl
@@ -1,5 +1,8 @@
 {{- define "repo.elasticsearch.config" -}}
 -Dindex.subsystem.name=elasticsearch
+{{- if eq (index .Values "alfresco-elasticsearch-connector" "elasticsearch" "protocol" | default .Values.global.elasticsearch.protocol) "https" }}
+-Delasticsearch.secureComms=https
+{{- end }}
 -Delasticsearch.host={{ index .Values "alfresco-elasticsearch-connector" "elasticsearch" "host" | default .Values.global.elasticsearch.host }}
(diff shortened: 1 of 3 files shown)


helm/alfresco-content-services/templates/_helpers-elasticsearch.tpl
```diff
@@ -1,5 +1,8 @@
 {{- define "repo.elasticsearch.config" -}}
 -Dindex.subsystem.name=elasticsearch
+{{- if eq (index .Values "alfresco-elasticsearch-connector" "elasticsearch" "protocol" | default .Values.global.elasticsearch.protocol) "https" }}
+-Delasticsearch.secureComms=https
+{{- end }}
```
aborroy, line 5: In addition to this environment variable, `encryption.ssl.truststore.*` values should include also the public certificate from the Elasticsearch server TLS protocol.

aborroy (changes requested):
Adding truststore configuration.

helm/alfresco-content-services/templates/_helpers-elasticsearch.tpl
gionn: why is that needed? our images should have proper cacerts and ssl certificates used by aws opensearch endpoints should already be recognized as trusted

aborroy (approved):
LGTM

helm/alfresco-content-services/templates/_helpers-elasticsearch.tpl
aborroy: You're right @gionn 
Repository is using a custom truststore to communicate with SOLR, but for Elasticsearch standard REST Client is used (so default JVM `cacerts` is enough).

PR #796, opened 2022-10-26
OPSEXP-1811: fix alf-ES config

alxgomz:
Ref: OPSEXP-1811

### DIFF
--- a/helm/alfresco-content-services/tests/alfresco-common_test.yaml
+++ b/helm/alfresco-content-services/tests/alfresco-common_test.yaml
@@ -19,11 +19,13 @@ tests:
       - matchRegex:
           path: metadata.name
           pattern: ^RELEASE-NAME-alfresco-cs-
+        template: ingress-repository.yaml
   - it: should render with value set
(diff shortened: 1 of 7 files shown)


helm/alfresco-content-services/tests/alfresco-common_test.yaml
```diff
@@ -19,11 +19,13 @@ tests:
       - matchRegex:
           path: metadata.name
           pattern: ^RELEASE-NAME-alfresco-cs-
+        template: ingress-repository.yaml
```
gionn, line 22: I remember to have read that the default template is the first one listed in templates, so there is no need for this?

alxgomz: no really but I realized that the plugin we use now is not the most maintained one. The new one I'd like we use later on (once we start moving to individual charts) has a different behavior (it defaults to the whole list and expect each template to render the asserted path.
So specifying the template works in both cases

gionn (approved):

PR #797, opened 2022-10-26
OPSEXP-1812 Fix for missing trustAnchors in repository deployment

gionn:
```
2022-10-14 14:40:09,598 ERROR [org.alfresco.repo.search.impl.elasticsearch.contentmodelsync.ElasticsearchIndexServic
e] [elasticsearch-initializer] Failed to check if basic mapping is loaded on alfresco
java.io.IOException: java.security.InvalidAlgorithmParameterException: the trustAnchors parameter must be non-empty
        at org.elasticsearch.client.RestClient.extractAndWrapCause(RestClient.java:884)
```

OPSEXP-1812

### DIFF
--- a/helm/alfresco-content-services/templates/config-repository.yaml
+++ b/helm/alfresco-content-services/templates/config-repository.yaml
@@ -154,3 +154,4 @@ data:
       -Dmessaging.broker.url=$BROKER_URL
       -Dmessaging.broker.username=$BROKER_USERNAME
       -Dmessaging.broker.password=$BROKER_PASSWORD
+      -Dencryption.ssl.truststore.location=/etc/pki/java/cacerts
--- a/helm/alfresco-content-services/templates/config-repository.yaml
+++ b/helm/alfresco-content-services/templates/config-repository.yaml
@@ -154,4 +154,4 @@ data:
       -Dmessaging.broker.url=$BROKER_URL
       -Dmessaging.broker.username=$BROKER_USERNAME
       -Dmessaging.broker.password=$BROKER_PASSWORD
-      -Dencryption.ssl.truststore.location=/etc/pki/java/cacerts
+      -Dencryption.ssl.truststore.location=$JAVA_HOME/lib/security/cacerts
--- a/helm/alfresco-content-services/tests/repository_test.yaml
+++ b/helm/alfresco-content-services/tests/repository_test.yaml
@@ -14,9 +14,7 @@ tests:
       - equal:
           path: metadata.name
           value: RELEASE-NAME-alfresco-cs-repository
-      - equal:
-          path: spec.template.metadata.annotations.checksum/config
-          value: 65b89491b06b9ccd8146424a8eab9d7a9d109a18db722e1276f1223ee37fd47c
+        template: deployment-repository.yaml
   - it: should have basic metadata in place
     values: *testvalues
     asserts:
@@ -26,5 +24,5 @@ tests:
         template: config-repository.yaml
       - matchRegex:
           path: data.CATALINA_OPTS
-          pattern: -Dencryption.ssl.truststore.location=/.*
+          pattern: -Dencryption.ssl.truststore.location=\$JAVA_HOME/lib/security/cacerts
         template: config-repository.yaml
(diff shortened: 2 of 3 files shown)


helm/alfresco-content-services/templates/config-repository.yaml
```diff
@@ -154,3 +154,4 @@ data:
       -Dmessaging.broker.url=$BROKER_URL
       -Dmessaging.broker.username=$BROKER_USERNAME
       -Dmessaging.broker.password=$BROKER_PASSWORD
+      -Dencryption.ssl.truststore.location=/etc/pki/java/cacerts
```
gionn, line 157: still wondering why it's needed given that `/usr/lib/jvm/java/lib/security/cacerts` is already a symlink to `/etc/pki/java/cacerts`

alxgomz: becuase this `encryption.ssl.truststore.*` are actually linked to an Alfresco implementation of keystore which doesn't default to the JVM keystores:
https://github.com/Alfresco/alfresco-enterprise-repo/blob/master/repository/src/main/java/org/alfresco/repo/search/impl/elasticsearch/client/ElasticsearchHttpClientFactory.java#L83

helm/alfresco-content-services/tests/repository_test.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - config-repository.yaml
+  - secret-database.yaml
+  - secret-s3.yaml
+  - config-dev-log4j-properties.yaml
+tests:
+  - it: should have basic metadata in place
+    values: &testvalues
+      - values/test_values.yaml
+    asserts:
+      - equal:
```
alxgomz, line 14: please specify the actual template to be asserted in each assertion, as defaults seem to change with plugin implementation

helm/alfresco-content-services/templates/config-repository.yaml
alxgomz: Other than that, we should give a path relative to JAVA_HOME so it doesn't break if we change JDK distribution or OS

helm/alfresco-content-services/tests/repository_test.yaml
```diff
@@ -0,0 +1,30 @@
...
+          path: metadata.name
+          value: RELEASE-NAME-alfresco-cs-repository
+      - equal:
+          path: spec.template.metadata.annotations.checksum/config
+          value: 65b89491b06b9ccd8146424a8eab9d7a9d109a18db722e1276f1223ee37fd47c
+  - it: should have basic metadata in place
+    values: *testvalues
+    asserts:
+      - equal:
```
alxgomz, line 23: IIRC this is already asserted in the alfresco-common suite

helm/alfresco-content-services/tests/repository_test.yaml
```diff
@@ -0,0 +1,30 @@
...
+          path: metadata.name
+          value: RELEASE-NAME-alfresco-cs-repository-configmap
+        template: config-repository.yaml
+      - matchRegex:
+          path: data.CATALINA_OPTS
+          pattern: -Dencryption.ssl.truststore.location=/.*
```
alxgomz, line 29: Can you also handle spaces and avoid using wildcards (we have a static string). See `search_test.yaml`.
Maybe also add an test where custom metadata keystore is used: https://github.com/Alfresco/acs-deployment/blob/ecbe4090e450cc7477a634e53f419b11987d3fbf/docs/helm/examples/custom-metadata-keystore.md

alxgomz (changes requested):

helm/alfresco-content-services/tests/repository_test.yaml
gionn: it's more to have a very basic assertion in every test to make sure that suite is setup correctly, we could remove all of them once we have more tests

helm/alfresco-content-services/tests/repository_test.yaml
gionn: rewritten it.

don't want to test custom metadata keystore given that logic is really simple (values plus a default) and that is out of scope for this pr

alxgomz (approved):

PR #803, opened 2022-11-04
OPSEXP-1805 Make an url builder function to be dry in repo and share cm

gionn:
OPSEXP-1805

### DIFF
--- a/helm/alfresco-content-services/templates/_helpers-url.tpl
+++ b/helm/alfresco-content-services/templates/_helpers-url.tpl
@@ -1,3 +1,23 @@
+{{- define "alf.protocol" -}}
+{{ tpl (.Values.externalProtocol | default "http") $ }}
+{{- end -}}
+
+{{- define "alf.repo.host" -}}
+{{ tpl (.Values.externalHost | default (printf "%s-repository" (include "content-services.shortname" .))) $ }}
+{{- end -}}
+
+{{- define "alf.repo.port" -}}
+{{ tpl ((.Values.externalPort | default .Values.repository.service.externalPort) | toString) $ }}
+{{- end -}}
+
+{{- define "alf.share.host" -}}
+{{ tpl (.Values.externalHost | default (printf "%s-share" (include "content-services.shortname" .))) $ }}
+{{- end -}}
+
+{{- define "alf.share.port" -}}
+{{ tpl (.Values.externalPort | default .Values.share.service.externalPort | toString) $ }}
+{{- end -}}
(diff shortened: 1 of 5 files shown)


helm/alfresco-content-services/templates/_helpers-url.tpl
```diff
@@ -0,0 +1,27 @@
+{{- define "alf.protocol" -}}
+{{ tpl (.Values.externalProtocol | default "http") $ }}
```
alxgomz, line 2: Do you think we really need to use `tpl`  function? It looks to me all we do here is regular value expansion.

helm/alfresco-content-services/templates/_helpers-url.tpl
```diff
@@ -0,0 +1,27 @@
...
+{{- define "alf.share.host" -}}
+{{ tpl (.Values.externalHost | default (printf "%s-share" (include "content-services.shortname" .))) $ }}
+{{- end -}}
+
+{{- define "alf.share.port" -}}
+{{ tpl (.Values.externalPort | default .Values.share.service.externalPort | toString) $ }}
+{{- end -}}
+
+{{- define "url.builder" -}}
+{{- if or (and (eq .protocol "https") (eq .port "443")) (and (eq .protocol "http") (eq .port "80")) -}}
```
alxgomz, line 22: You're using arguments for this named template but those argument are actually other named templates contained in this very same file. Should it be better to simply call those named templates in this one so we can fully get rid of the `$alf*` variables?

helm/alfresco-content-services/tests/repository_test.yaml
```diff
@@ -7,22 +7,50 @@ templates:
...
       - matchRegex:
           path: data.CATALINA_OPTS
           pattern: -Dencryption.ssl.truststore.location=\$JAVA_HOME/lib/security/cacerts
         template: config-repository.yaml
+  - it: should have configmap with ALFRESCO_OPTS urls without ports when are the same of the protocol default
+    values: *testvalues
+    asserts:
+      - matchRegex:
+          path: data.ALFRESCO_OPTS
+          pattern: -Daos.baseUrlOverwrite=http://RELEASE-NAME-alfresco-cs-repository/alfresco/aos
```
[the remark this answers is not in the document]
alxgomz: don't like it? :)

alxgomz (changes requested):

alxgomz (approved):

PR #812, opened 2022-11-16
OPSEXP-1796 Prepare chart versions for next iteration

gionn:
OPSEXP-1796


alxgomz (commented):
If I understand correctly the first publish try will fails on content-services chart as common will most likely be missing. Is the intention to just retry?

gionn:
> If I understand correctly the first publish try will fails on content-services chart as common will most likely be missing. Is the intention to just retry?

no, pre-commit and helm tests are fine thanks to use-local-deps, while publish of each chart is done sequentially and in the expected order. Apart from temporary glitches (like caching on github cdn side) everything should be fine.

alxgomz:
yes, when I say "likely to be missing" I'm referring to the caching behavior we saw yesterday.

alxgomz (approved):

PR #814, opened 2022-11-21
OPSEXP-1585: split Transformation filestore persistence

alxgomz:
Ref: OPSEXP-1585
Make filestore persistent volume independant by default


helm/alfresco-content-services/templates/_helpers-persistence.tpl
```diff
@@ -0,0 +1,36 @@
...
+  {{- end }}
+  {{- if .storageClass.name }}
+  storageClassName: {{ .storageClass.name }}
+  {{- end }}
+  {{- end }}
+  resources:
+    requests:
+      storage: {{ .baseSize | default "20Gi" }}
+{{- end }}
+{{- end -}}
```
gionn, line 36: should be good to have in alfresco-common?

so we can use it also in aps chart

helm/alfresco-content-services/templates/deployment-filestore.yaml
```diff
@@ -48,6 +48,10 @@ spec:
                   values:
                   - {{ template "alfresco.shortname" . }}-filestore
               topologyKey: kubernetes.io/hostname
+      securityContext:
+        runAsUser: 33030
+        runAsGroup: 1000
+        fsGroup: 1000
```
gionn, line 54: why you need to override those?

helm/alfresco-content-services/values.yaml
```diff
@@ -466,6 +466,21 @@ filestore:
...
+    # -- Persist filestore data
+    enabled: true
+    # -- Specify a storageClass for dynamic provisioning
+    storageClass:
+      enabled: true
+      accessModes:
+        - ReadWriteOnce
+      # -- Custom storage class name
+      name: null
+    # -- Use pre-provisioned volume (through claim) instead of dynamiv provisioning
```
gionn, line 479:
```suggestion
    # -- Use pre-provisioned volume (through claim) instead of dynamic provisioning
```

helm/alfresco-content-services/values.yaml
```diff
@@ -466,6 +466,21 @@ filestore:
...
+    existingClaim: null
+    data:
+      mountPath: "/tmp/Alfresco"
+      subPath: "alfresco-content-services/filestore-data"
```
gionn, line 483: this is the first one but the object is to have different storage for each component, so may be worth adding all of those under a common filestore key?

helm/alfresco-content-services/templates/_helpers-persistence.tpl
alxgomz: Of course yes. But for now it'll be only used in the main chart. The one who needs it move it to alfresco-common ;)

helm/alfresco-content-services/values.yaml
alxgomz: current chart already have `.persistence.component` keys. Here we move to a model where each component embeds its own persistence config in `component.persistence`

helm/alfresco-content-services/templates/_helpers-persistence.tpl
gionn: I guess who will be then... =D

gionn (changes requested):

helm/alfresco-content-services/values.yaml
```diff
@@ -466,6 +466,21 @@ filestore:
...
+  persistence:
+    # -- Persist filestore data
+    enabled: true
+    # -- Specify a storageClass for dynamic provisioning
+    storageClass:
+      enabled: true
+      accessModes:
+        - ReadWriteOnce
+      # -- Custom storage class name
+      name: null
```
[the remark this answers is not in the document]
alxgomz: I have actually changed the structure of the `persistence` object to match what bitnami does

gionn (approved):

PR #826, opened 2022-12-12
OPSEXP-1875: exclude actual go templates

alxgomz:
Ref OPSEXP-1875

Addiotnal PR to introduce updatecli config in acs-deplyment (as beta considering opened updatecli issues (https://github.com/updatecli/updatecli/issues/1028 & https://github.com/updatecli/updatecli/issues/794)


gionn (changes requested):
I would move this to a new task specific for updatecli support in acs-deployment

PR #828, opened 2022-12-13
OPSEXP-1875 Prepare for the next iteration

gionn:
OPSEXP-1875


alxgomz (changes requested):
Would you mind setting the subcharts' Chart.yml file as well (those might not need to be set to SNAPSHOT version for now

alxgomz (dismissed):

PR #829, opened 2022-12-13
OPSEXP-1901 Document manual release steps - waiting for automation

gionn:
OPSEXP-1901


alxgomz (commented):
We don't have any use of the -SNAPSHOT tag for the repo. I would remove it.

gionn:
> We don't have any use of the -SNAPSHOT tag for the repo. I would remove it.

we can't, it triggers the publishing to the incubator repo that is mandatory to not break dependencies resolution outside of the GHA workflow.

alxgomz:
true. I forgot about it... up to the point I needed it :)

alxgomz (approved):

PR #833, opened 2023-01-03
OPSEXP-1651 Use existing secret to configure activemq credentials in embedded chart

gionn:
OPSEXP-1651

- [x] support existing secret for embedded activemq
- [x] add unit tests

[OPSEXP-1651]: https://alfresco.atlassian.net/browse/OPSEXP-1651?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/helm/alfresco-content-services/charts/activemq/Chart.yaml
+++ b/helm/alfresco-content-services/charts/activemq/Chart.yaml
@@ -1,7 +1,7 @@
 ---
 apiVersion: v2
 appVersion: 5.16.4
-description: A Helm chart Providing Apache ActiveMQ.
+description: A Helm chart providing a basic Apache ActiveMQ deployment required to evaluate ACS (do not use in production).
 keywords:
   - activemq
   - message broker
--- a/helm/alfresco-content-services/charts/activemq/Chart.yaml
+++ b/helm/alfresco-content-services/charts/activemq/Chart.yaml
@@ -1,6 +1,5 @@
 ---
 apiVersion: v2
-appVersion: 5.16.4
 appVersion: 5.17.1
 description: A Helm chart providing a basic Apache ActiveMQ deployment required to evaluate ACS (do not use in production).
 keywords:
--- a/helm/alfresco-content-services/charts/activemq/templates/deployment-activemq.yaml
+++ b/helm/alfresco-content-services/charts/activemq/templates/deployment-activemq.yaml
@@ -39,9 +39,15 @@ spec:
         - name: ACTIVEMQ_BROKER_NAME
           value: "{{ template "activemq.fullname" . }}"
         - name: ACTIVEMQ_ADMIN_LOGIN
-          value: "{{ .Values.adminUser.username }}"
+          valueFrom:
+            secretKeyRef:
+              name: {{ default (printf "%s-brokersecret" (include "content-services.shortname" $)) $.Values.adminUser.existingSecretName }}
+              key: BROKER_USERNAME
         - name: ACTIVEMQ_ADMIN_PASSWORD
-          value: "{{ .Values.adminUser.password }}"
+          valueFrom:
+            secretKeyRef:
+              name: {{ default (printf "%s-brokersecret" (include "content-services.shortname" $)) $.Values.adminUser.existingSecretName }}
+              key: BROKER_PASSWORD
         ports:
         - name: stomp
           containerPort: {{ .Values.services.broker.ports.internal.stomp | default 61613 }}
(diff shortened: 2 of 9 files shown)


helm/alfresco-content-services/templates/secret-message-broker.yaml
```diff
@@ -9,7 +9,7 @@ type: Opaque
 data:
   {{- if .Values.activemq.enabled }}
   BROKER_URL: {{ printf "failover:(nio://%s-broker:61616)?timeout=3000&jms.useCompression=true" (include "activemq.fullname" .) | b64enc | quote }}
-  BROKER_USERNAME: {{ .Values.activemq.adminUser.username | b64enc | quote }}
+  BROKER_USERNAME: {{ .Values.activemq.adminUser.user | b64enc | quote }}
```
gionn, line 12: to uniform with messageBroker

gionn (commented):
@alxgomz please review the approach before I start writing unit tests

helm/alfresco-content-services/values.yaml
```diff
@@ -542,6 +542,11 @@ activemq:
   # -- Possibility to choose Node for pod, with a key-value pair label
   # e.g {"kubernetes.io/hostname": multinode-demo-m02}
   nodeSelector: {}
+  adminUser:
+    # -- Default username for the embedded broker admin user
+    user: admin
+    # -- Default password for the embedded broker admin user
+    password: admin
```
alxgomz, line 549: We're already using a `messageBroker` data object for clients. I think we should not have different config elements for clients and services. 
Either we should create a secret as part of the main charts and use it as an `existingSecret`, or set credentials in values and find a way to share them across the other components.
At first sight the former looks like a better approach to me but I'm not sure this is really what best suits a reference chart

alxgomz (changes requested):

helm/alfresco-content-services/charts/activemq/templates/deployment-activemq.yaml
```diff
@@ -39,9 +39,15 @@ spec:
...
+          valueFrom:
+            secretKeyRef:
+              name: {{ default (printf "%s-brokersecret" (include "content-services.shortname" $)) $.Values.adminUser.existingSecretName }}
+              key: BROKER_USERNAME
         - name: ACTIVEMQ_ADMIN_PASSWORD
-          value: "{{ .Values.adminUser.password }}"
+          valueFrom:
+            secretKeyRef:
+              name: {{ default (printf "%s-brokersecret" (include "content-services.shortname" $)) $.Values.adminUser.existingSecretName }}
+              key: BROKER_PASSWORD
```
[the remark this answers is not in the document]
gionn: only when secret keys and env vars match, in this case they need to be overridden and env/valueFrom should be fine

helm/alfresco-content-services/charts/activemq/values.yaml
```diff
@@ -7,8 +7,8 @@ image:
   tag: 5.17.1-jre11-rockylinux8
   pullPolicy: IfNotPresent
 adminUser:
-  username: admin
-  password: admin
```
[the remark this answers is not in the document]
alxgomz: Still, I think it's better to keep it in the subchart doc, even with no value, just for doc really.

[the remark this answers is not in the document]
gionn: I have updated the README template of the activemq chart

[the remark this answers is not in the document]
alxgomz: Let's meet quickly tomorrow because I really don't understand why you're doing that. I must have missed something.

alxgomz (dismissed):

helm/alfresco-content-services/charts/activemq/Chart.yaml
```diff
@@ -1,7 +1,8 @@
 ---
 apiVersion: v2
+appVersion: 5.16.4
```
gionn, line 3:
```suggestion
```

helm/alfresco-content-services/templates/secret-message-broker.yaml
mteodori, line 12: we always use username as variable name otherwise here it should have been just name but not user

gionn: generally speaking I would have preferred `username` but we have `messageBroker.user` in the main chart and is spread in different subchart so I preferred to be consistent and hopefully switch to `username` everywhere at a later time

PR #837, opened 2023-01-04
OPSEXP-1651 Use secret to configure elasticsearch credentials

gionn:
Ref: OPSEXP-1651

- [x] elasticsearch credentials in existing secret
- [x] add unit tests

[OPSEXP-1651]: https://alfresco.atlassian.net/browse/OPSEXP-1651?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/helm/alfresco-content-services/charts/alfresco-elasticsearch-connector/templates/secret-elasticsearch.yaml
+++ b/helm/alfresco-content-services/charts/alfresco-elasticsearch-connector/templates/secret-elasticsearch.yaml
@@ -10,6 +10,6 @@ metadata:
     component: search
 type: Opaque
 data:
-  ELASTICSEARCH_USERNAME: {{ .Values.elasticsearch.user | default .Values.global.elasticsearch.user | b64enc | quote }}
-  ELASTICSEARCH_PASSWORD: {{ .Values.elasticsearch.password  | default .Values.global.elasticsearch.password | b64enc | quote }}
+  ELASTICSEARCH_USERNAME: {{ .Values.elasticsearch.user | default .Values.global.elasticsearch.user | default "" | b64enc | quote }}
+  ELASTICSEARCH_PASSWORD: {{ .Values.elasticsearch.password | default .Values.global.elasticsearch.password | default "" | b64enc | quote }}
 {{- end }}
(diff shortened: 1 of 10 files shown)


helm/alfresco-content-services/charts/alfresco-elasticsearch-connector/templates/secret-elasticsearch.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if not .Values.global.elasticsearch.existingSecretName }}
```
alxgomz, line 1: Why is the `existingSecret` only provided through the global scope? We usually use global as a fallback for the root scope in case the values isn't available (and as a side effect provides convenient way of spreading the values)

gionn: took a shortcut and forgot to fix it later, thanks for pointing it

helm/alfresco-content-services/charts/alfresco-elasticsearch-connector/tests/secret-elasticsearch_test.yaml
```diff
@@ -0,0 +1,47 @@
...
+    set:
+      global:
+        elasticsearch:
+          user: admin
+          password: letmein
+    asserts:
+      - isNotEmpty:
+          path: data.ELASTICSEARCH_USERNAME
+      - isNotEmpty:
+          path: data.ELASTICSEARCH_PASSWORD
```
alxgomz, line 30: why not testing for the resulting b64 str?

alxgomz (dismissed):

helm/alfresco-content-services/charts/alfresco-elasticsearch-connector/tests/secret-elasticsearch_test.yaml
gionn: typical issue of doing the same thing in two different places in two different days 😀

alxgomz (approved):

PR #842, opened 2023-01-10
OPSEXP-1651 Use existing secret to configure repo admin password

gionn:
OPSEXP-1651

### DIFF
--- a/helm/alfresco-content-services/templates/deployment-repository.yaml
+++ b/helm/alfresco-content-services/templates/deployment-repository.yaml
@@ -55,12 +55,9 @@ spec:
               name: {{ default (printf "%s-repository-secret" (include "content-services.shortname" .)) .Values.repository.existingSecretName }}
           - secretRef:
               name: {{ default (printf "%s-dbsecret" (include "content-services.shortname" $)) $.Values.postgresql.existingSecretName }}
-          {{- if .Values.s3connector.enabled }}
-          {{- if .Values.s3connector.secrets }}
+          {{- if and .Values.s3connector.enabled .Values.s3connector.secrets }}
           - secretRef:
-              # s3 connector secret keys
               name: {{ template "content-services.shortname" . }}-s3secret
-          {{- end }}
           {{- end }}
           {{- if .Values.email.server.enabled }}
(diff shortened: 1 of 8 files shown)


helm/alfresco-content-services/templates/deployment-repository.yaml
```diff
@@ -24,6 +24,7 @@ spec:
         checksum/config: {{ include (print $.Template.BasePath "/config-repository.yaml") . | sha256sum }}
         checksum/secretDatabase: {{ include (print $.Template.BasePath "/secret-database.yaml") . | sha256sum }}
         checksum/secretS3: {{ include (print $.Template.BasePath "/secret-s3.yaml") . | sha256sum }}
+        checksum/secret-repository: {{ include (print $.Template.BasePath "/secret-repository.yaml") . | sha256sum }}
```
alxgomz, line 27: Did you tests that? I'm curious to know whether this still works at all.
Also, we don't want to trigger redeploy when some one changes the admin password as it is anyway just used at bootstrap

alxgomz (changes requested):

helm/alfresco-content-services/templates/deployment-repository.yaml
gionn: oh that's true, but I added it BEFORE knowing that password is set only at bootstrap :D

more in general, yes it works as expected, and I was thinking to create a new issue to make sure that at least the cm/secret that involves most interesting or most commonly changed values are handled in this way.

gionn, line 27:
```suggestion
```

alxgomz: Agreed. We'll however need to be very cautious about what trigger redeployment as the repo deployment policy is not set to `RollingUpdate` but instead to `Recreate`?
Maybe we should investigate switching to rolling updates first...?

upandey123 (dismissed):

upandey123:
the branch has conflicts that needs to be addressed.

alxgomz (approved):

PR #844, opened 2023-01-11
OPSEXP-1651 Support existing secret for s3 connector

gionn:
Ref: OPSEXP-1651

- [x] add test in deployment-repository


helm/alfresco-content-services/tests/secret-s3_test.yaml
```diff
@@ -0,0 +1,45 @@
+---
+suite: test s3 credentials secret manifest
+templates:
+  - secret-s3.yaml
+tests:
+  - it: should have empty credentials as default
```
alxgomz, line 6: Can you rename it? I find it a. bit miss-leading as there could also be a secret rendered with empty credentials using the config bellow:
```yaml
    s3connector:
      enabled: true
      config:
        bucketName: somebucket
        bucketLocation: someregion
      secrets:
        accessKey: null
        secretKey: null
```

This is still a valid config if you prefer setting up S3 access using IAM policies

alxgomz (dismissed):

PR #845, opened 2023-01-11
OPSEXP-1651 Support existing secret for mail password

gionn:
Ref: OPSEXP-1651

- [x] add test in deployment-repository


alxgomz (approved):
Worth working on OPSEXP-1826 after that as `email`  & `mail` are kkind of mixed up here

PR #847, opened 2023-01-13
OPSEXP-1925: update doc to reflect persistance handling changes

alxgomz:
Ref: OPSEXP-1925
Also fixes: OPSEXP-1879


docs/helm/examples/with-ai.md
```diff
@@ -23,9 +23,9 @@ helm install acs alfresco/alfresco-content-services \
 --set externalPort="443" \
 --set externalProtocol="https" \
 --set externalHost="acs.YOUR-DOMAIN-NAME" \
---set persistence.enabled=true \
---set persistence.storageClass.enabled=true \
---set persistence.storageClass.name="nfs-client" \
+--set repository.persistence.enabled=false \
```
gionn, line 26: why false? if it's a required we could briefly explain it because I wasn't expecting that

gionn (approved):
aws service diagram looks a bit too crowded but I am not sure if just trying a different arrangement would help

gionn (approved):

PR #850, opened 2023-01-16
OPSEXP-1879 Enforce markdown linting

gionn:
Ref: OPSEXP-1879


alxgomz (approved):
Sorry I thought it was in my previous PR but actually missed it

PR #859, opened 2023-01-20
OPSEXP-819 Have cpu resources set everywhere

gionn:
Ref: OPSEXP-819

to do:
- [x] fix unit tests after first round of reviews


helm/alfresco-content-services/charts/alfresco-elasticsearch-connector/values.yaml
```diff
@@ -11,9 +11,11 @@ tolerations: []
 affinity: {}
 resources:
   requests:
-    memory: 256Mi
+    cpu: "0.5"
+    memory: "256Mi"
   limits:
-    memory: 2048Mi
+    cpu: "4"
```
alxgomz, line 17: 4 is a rather high value. I understand indexing process may need that but does the mediation pod also needs that?

helm/alfresco-content-services/charts/alfresco-elasticsearch-connector/values.yaml
```diff
@@ -60,6 +62,19 @@ reindexing:
...
+      cpu: "0.5"
+      memory: "128Mi"
+    limits:
+      cpu: "2"
+      memory: "512Mi"
+  initcontainers:
+    waitForRepository:
+      resources:
+        limits:
+          cpu: "0.5"
```
alxgomz, line 76: that sounds a lot for a container that simply curl the repo. Can we lower it?

helm/alfresco-content-services/values.yaml
```diff
@@ -145,8 +146,10 @@ ooiService:
     runAsUser: 33006
   resources:
     requests:
+      cpu: "0.25"
       memory: "1000Mi"
```
alxgomz, line 150: not worth lowering?

helm/alfresco-content-services/values.yaml
```diff
@@ -184,8 +187,10 @@ msTeamsService:
     runAsUser: 33041
   resources:
     requests:
+      cpu: "0.5"
       memory: "1000Mi"
```
alxgomz, line 191: not worth lowering?

helm/alfresco-content-services/values.yaml
```diff
@@ -299,8 +308,10 @@ imagemagick:
     runAsUser: 33002
   resources:
     requests:
-      memory: "1000Mi"
+      cpu: "0.25"
```
alxgomz, line 311: sounds really low to me as a default for a process that's rather cpu intensive

helm/alfresco-content-services/values.yaml
```diff
@@ -334,8 +345,10 @@ libreoffice:
     externalPort: 80
   resources:
     requests:
-      memory: "1000Mi"
+      cpu: "0.25"
```
alxgomz, line 348: same. I would double it or even set to 1  cpu this one and the imagemagick one

test/community-integration-test-values.yaml
```diff
@@ -2,6 +2,7 @@
 repository:
   resources:
     requests:
+      cpu: "0.01"
```
gionn, line 5: on GHA we only have 2 cpu to run the community tests so I put here the minimum value to make it work and still having a value to be sure that are really handled

helm/alfresco-content-services/charts/alfresco-elasticsearch-connector/values.yaml
alxgomz: yes for now let's move on as is
OPSEXP-1976

alxgomz (approved):

PR #862, opened 2023-01-24
OPSEXP-1931 Create a dedicated workflow for helm unit tests

slohe1:


.github/workflows/helm-community-test.yml
```diff
@@ -0,0 +1,52 @@
+---
+name: Helm (Community-unit-test)
```
alxgomz, line 2: The workflow is not community specific. It happened to live in the community workflow before but it actually applies to enterprise as well.
Remove community mentions and rename workflow file

.github/workflows/helm-community-test.yml
```diff
@@ -0,0 +1,52 @@
...
+    branches:
+      - 'master'
+    paths:
+      - 'helm/**'
+      - 'test/postman/helm/**'
+      - '.github/workflows/helm*'
+  push:
+    branches:
+      - 'master'
+      - pr-*
```
alxgomz, line 14: looks like a legcay stuff... What do you think about getting rid of it?

alxgomz (changes requested):

gionn (approved):

alxgomz (approved):

PR #863, opened 2023-01-25
OPSEXP-1938 Avoid using JAVA_TOOL_OPTIONS in repository for setting solr secret

gionn:
So the secret is not getting printed into the logs of repo at startup

Ref: OPSEXP-1938


alxgomz (commented):
I’m not confortable with using the variable in the command line argument. It’ll show up in a simple ps output. Did you try using regular JAVA_OPTS? I don’t think this one creates a log line on startup

alxgomz (changes requested):

PR #870, opened 2023-01-31
OPSEXP-1958 Introduce docs to handle upgrades

gionn:
Documenting the new cpu resources breaking change and move the existing solr tracking and persistence upgrades.

Ref: OPSEXP-1958 && OPSEXP-819


docs/helm/upgrades.md
```diff
@@ -0,0 +1,99 @@
+# Upgrading the helm charts
+
+Our helm charts are continuously improved and sometimes arise the need to introduce a breaking change.
+
+To get an overview of the changes in each release, first take a look at the
+release notes that are available via [GitHub Releases](https://github.com/Alfresco/acs-deployment/releases).
+
+Here follows a more detailed explaination of any breaking change grouped by version in which they have been released.
+
+## To be released
```
gionn, line 10: this implies that part of the release process should be to rename this section to the current release and make a new to be released section for the future (only for GA release or also Mx?)

docs/helm/README.md
```diff
@@ -65,6 +65,12 @@ the installation.
 
 For the best results we recommend [deploying ACS to AWS EKS](./eks-deployment.md). If you have a machine with at least 16GB of memory you can also [deploy using Docker for Desktop](./docker-desktop-deployment.md).
 
+The recommended cluster resources for the Enterprise version are at least 3
+nodes with 12 cpu cores and 32 GB of memory in total. You can install with lower
```
alxgomz, line 69: Mentions that this is for the components enabled by default (which users may want to disable, or enable new ones)

docs/helm/README.md
```diff
@@ -65,6 +65,12 @@ the installation.
...
+requirements by fine tuning the [resource requests](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#resource-units-in-kubernetes)
+available in the [values](../helm/../../helm/alfresco-content-services/values.yaml)
```
alxgomz, line 71:
```suggestion
available in the [values](../../helm/alfresco-content-services/values.yaml)
```

docs/helm/upgrades.md
alxgomz: @gionn IMHO the value justifies the burden. I would personnally keep everything under `To be released` (so M & A releases as well), but I know that customers are using the charts -M releases.
Though that could be a first step of aligning this repo's master branch with what we do in Ansible. @vprovaggi any inputs? 

@anxumalo @Amy-Hyland We're thinking about introducing this to the helm documentation. I guess that's something of interest to you?

docs/helm/README.md
gionn: looks like my autocomplete got completely mad lol

docs/helm/upgrades.md
gionn: I think it's reasonable to reference here also pre-release until we make the next GA. During the GA release process, we merge any -M* under the same GA version section.

docs/helm/upgrades.md
```diff
@@ -0,0 +1,99 @@
...
+    limits:
+      cpu: "4"
+      memory: "3000Mi"
+```
+
+That means that each `repository` pod will have a reservation of 1 full cpu core
+and enables bursting up to 4 cpu cores. You can also reduce the number of
+replicas for the components that allows it.
+
+## 5.3.0
```
alxgomz, line 43:
```suggestion
## 5.4.0-M1
```

alxgomz (changes requested):

docs/helm/upgrades.md
gionn: good catch, I probably grabbed the wrong ref in the blame of the previous doc

alxgomz (approved):

docs/helm/upgrades.md
anxumalo: > @anxumalo @Amy-Hyland We're thinking about introducing this to the helm documentation. I guess that's something of interest to you?

@alxgomz Oh yes, it'll make it easier to follow what's changed when, as the docs are prepared for the GA release. Thanks for the heads up!! 👍🏼

PR #874, opened 2023-02-01
OPSEXP-1353 Removed Alfresco Sync service HPA chart

upandey123:
Sync service is the only Chart that has some kind of autoscaling.

It is untested and leverage a now depreceated approach:

 

autoscaling/v2beta1 HorizontalPodAutoscaler is deprecated in v1.22+, unavailable in v1.25+; use autoscaling/v2beta2 HorizontalPodAutoscaler
Remove it from the subchart


helm/alfresco-content-services/charts/alfresco-sync-service/templates/hpa-syncservice.yaml
```diff
@@ -1,33 +0,0 @@
-{{- if .Values.syncservice.enabled }}
-{{- if .Values.syncservice.horizontalPodAutoscaling.enabled }}
```
alxgomz, line 2: These values need to be removed from the `values.yaml` file

alxgomz (changes requested):

helm/alfresco-content-services/charts/alfresco-sync-service/templates/hpa-syncservice.yaml
upandey123: Removed {{- if .Values.syncservice.enabled }} but syncservice.horizontalPodAutoscaling.enabled is not inside values file.

upandey123 (commented):
Removed {{- if .Values.syncservice.enabled }} but syncservice.horizontalPodAutoscaling.enabled is not inside values file.

alxgomz:
> Removed {{- if .Values.syncservice.enabled }} but syncservice.horizontalPodAutoscaling.enabled is not inside values file.

No, you're looking at the wrong file. do not edit the `helm/alfresco-content-services/values.yaml` instead you should edit `helm/alfresco-content-services/charts/alfresco-sync-service/templates/hpa-syncservice.yaml`

upandey123:
As checked from my end all the HPA resources have been removed from the sync service. Should be good!

helm/alfresco-content-services/charts/alfresco-sync-service/values.yaml
```diff
@@ -15,7 +15,6 @@ podSecurityContext:
   runAsNonRoot: true
   fsGroup: 1000
 syncservice:
-  enabled: true
```
gionn, line 18: it's still used in the [deployment](https://github.com/Alfresco/acs-deployment/blob/master/helm/alfresco-content-services/charts/alfresco-sync-service/templates/deployment-syncservice.yaml#L1)

gionn (changes requested):

gionn (approved):
LGTM

I will merge it since codacy is broken again.

PR #886, opened 2023-02-08
OPSEXP-1880 Test and document log4jv2 configuration in helm charts

slohe1:


docs/helm/README.md
```diff
@@ -184,6 +184,16 @@ repository:
...
+In addition with the Log4j we have the logging level with Log4j2, below is the example:
+
+```yaml
+repository:
+  ...
+  extraLogStatements:
+    logger.alfresco-repo-content-transform-TransformerDebug.name: org.alfresco.repo.content.transform.TransformerDebug
+    logger.alfresco-repo-content-transform-TransformerDebug.level: debug
+```
+
```
alxgomz, line 196: I would place the log4v2 example first as it's now the default.
Also remove the identityservice debug line so it's more visible that with log4v2 2 lines are required while one only for log4jv1

helm/alfresco-content-services/tests/deployment-repository_test.yaml
```diff
@@ -17,6 +17,18 @@ tests:
           value: RELEASE-NAME-alfresco-cs-repository
         template: deployment-repository.yaml
 
+  - it: Log4j & Log4j2 test
+    values: *testvalues
+    set:
+      repository.extraLogStatements.testkey: testvalue
```
alxgomz, line 23: needs to be yaml formatted

helm/alfresco-content-services/tests/deployment-repository_test.yaml
```diff
@@ -17,6 +17,18 @@ tests:
...
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[0].envFrom
+          content:
+            configMapRef:
+              name: RELEASE-NAME-alfresco-cs-repository-configmap
+        template: deployment-repository.yaml
```
alxgomz, line 30: You are testing the envFrom section contains a reference to the `RELEASE-NAME-alfresco-cs-repository-configmap` cm... That doesn't make sense to me.
You should be testing: 
* the volumemount exists
* the cm has the right entries  with the right values.

helm/alfresco-content-services/tests/deployment-repository_test.yaml
alxgomz: the volumemount has a subpath set ( to `dev-log4j.properties`) did you check that the resulting pod really has both the `dev-log4j.properties` & `dev-log4j2.properties` mounted?. I would expect it doesn't.

helm/alfresco-content-services/templates/config-dev-log4j-properties.yaml
```diff
@@ -9,6 +9,11 @@ metadata:
 data:
   dev-log4j.properties: |-
   {{- range $key, $val := .Values.repository.extraLogStatements }}
-   log4j.logger.{{ $key }}={{ tpl $val $ }}
+   log4j.logger.{{ $key }}={{ $val }}
+  {{- end }}
+  dev-log4j2.properties: |-
```
alxgomz, line 14: I would rename the cm and its entries from `dev-` to `custom-` as it's not really development oriented feature.

alxgomz (changes requested):
In addition to the review comments I would also rename the volume from `dev-log4j-properties-volume` to something more generic like `repository-logging-properties-volume`

docs/helm/README.md
```diff
@@ -174,7 +174,6 @@ repository:
   ...
   extraLogStatements:
     org.alfresco.repo.content.transform.TransformerDebug: debug
```
alxgomz, line 176: I think that class is an old one and may not be in the repo anymore
```suggestion
    org.alfresco.repo.security.sync: debug
```

alxgomz (dismissed):

alxgomz (approved):

PR #897, opened 2023-02-27
OPSEXP-1852 Migrate to new search enterprise chart

gionn:
Ref: OPSEXP-1852


alxgomz (dismissed):
LGTM

alxgomz (approved):
Looks exactly like the behavior we discussed. Seeing it in action now I think we may want to think about having a `.global` fallback as well for `exisitngSecretNames` at some point

PR #908, opened 2023-03-09
OPSEXP-2002 Fixup volume handling when using external database

### LINKED ISSUE #894: When database.external=true, the repository deployment is missing volumes
I tried deploying with Helm using an external database and got an error. Trying to inspect the generated yaml files (doing a dry-run) showed that the {name}-alfresco-cs-repository deployment was generated without any volume definition:

image

The problem seems to be in helm/alfresco-content-services/templates/deployment-repository.yaml, line 124. I've noticed that {{- if eq .Values.database.external false }} is closed at the end of the file, instead of being closed before the "volumes:" definition

gionn:
Ref: OPSEXP-2002
Fix #894

### DIFF
--- a/helm/alfresco-content-services/tests/deployment-repository_test.yaml
+++ b/helm/alfresco-content-services/tests/deployment-repository_test.yaml
@@ -196,7 +196,7 @@ tests:
               memory: "3000Mi"
         template: deployment-repository.yaml
 
-  - it: should have wait-db-ready initcontainer with external database
+  - it: should have wait-db-ready initcontainer with in-cluster db
     values: *testvalues
     set:
       database.external: false
--- a/helm/alfresco-content-services/tests/deployment-repository_test.yaml
+++ b/helm/alfresco-content-services/tests/deployment-repository_test.yaml
@@ -195,3 +195,59 @@ tests:
               cpu: "4"
               memory: "3000Mi"
         template: deployment-repository.yaml
+
+  - it: should have wait-db-ready initcontainer with external database
+    values: *testvalues
+    set:
+      database.external: false
+    asserts:
+      - equal:
+          path: spec.template.spec.initContainers[0].name
+          value: wait-db-ready
+        template: deployment-repository.yaml
+
+  - it: should not have wait-db-ready initcontainer with external database
+    values: *testvalues
+    set:
+      database.external: true
+    asserts:
+      - isEmpty:
+          path: spec.template.spec.initContainers
+        template: deployment-repository.yaml
+      - isNotEmpty:
+          path: spec.template.spec.volumes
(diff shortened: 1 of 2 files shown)


helm/alfresco-content-services/tests/deployment-repository_test.yaml
```diff
@@ -195,3 +195,59 @@ tests:
               cpu: "4"
               memory: "3000Mi"
         template: deployment-repository.yaml
+
+  - it: should have wait-db-ready initcontainer with external database
```
alxgomz, line 199:
```suggestion
  - it: should have wait-db-ready initcontainer with in-cluster db
```

helm/alfresco-content-services/tests/deployment-repository_test.yaml
```diff
@@ -195,3 +195,59 @@ tests:
...
+          path: spec.template.spec.initContainers[0].name
+          value: wait-db-ready
+        template: deployment-repository.yaml
+
+  - it: should not have wait-db-ready initcontainer with external database
+    values: *testvalues
+    set:
+      database.external: true
+    asserts:
+      - isEmpty:
```
alxgomz, line 214: This test will break if we add an initContainer later on. Can't [`notContains`](https://github.com/helm-unittest/helm-unittest/blob/main/DOCUMENT.md#assertion-types) fulfill this check?

helm/alfresco-content-services/tests/deployment-repository_test.yaml
```diff
@@ -195,3 +195,59 @@ tests:
...
+          path: spec.template.spec.initContainers
+        template: deployment-repository.yaml
+      - isNotEmpty:
+          path: spec.template.spec.volumes
+        template: deployment-repository.yaml
```
alxgomz, line 219: Looks unrelated to the test or am I missing something?

alxgomz (changes requested):

helm/alfresco-content-services/tests/deployment-repository_test.yaml
gionn: it is to check explicitly for an effect that the bug was causing (empty volumes)

helm/alfresco-content-services/tests/deployment-repository_test.yaml
gionn: `contains` assertions works great with simple objects, but in this case we have a complex object and in order to pass I have to provide the full specs that in case of an initcontainer are pretty big (e.g. `command`) and I can't provide a subset of them (e.g. just check the name of the container).

I decided to go with isEmpty but indeed I am not satisfied with it

PR #913, opened 2023-03-22
OPSEXP-2028 Update development version

slohe1:
OPSEXP-2028

### DIFF
--- a/helm/alfresco-content-services/charts/alfresco-search/Chart.yaml
+++ b/helm/alfresco-content-services/charts/alfresco-search/Chart.yaml
@@ -13,11 +13,11 @@ keywords:
 name: alfresco-search
 sources:
   - https://github.com/Alfresco/acs-deployment
-version: 1.2.0-SNAPSHOT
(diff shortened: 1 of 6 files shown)


helm/alfresco-content-services/charts/alfresco-search/Chart.yaml
```diff
@@ -13,11 +13,11 @@ keywords:
 name: alfresco-search
 sources:
   - https://github.com/Alfresco/acs-deployment
-version: 1.2.0
+version: 1.2.0-SNAPSHOT
```
gionn, line 16: dependencies should be increased to next snapshot (1.3.0-SNAPSHOT) - check if the docs are explicit about this

gionn (changes requested):

slohe1 (commented):
@alxgomz @gionn  ```v5.4.0-SNAPSHOT``` this tag is already there. Do we make minor version change here?

gionn (approved):

PR #920, opened 2023-05-03
OPSEXP-2013 fixup postgresql resources example

gionn:
Ref: OPSEXP-2013


alxgomz (commented):
Can you please link it to a jira?

PR #929, opened 2023-05-18
OPSEXP-2109: update updatecli

alxgomz:
Updatecli pipeline and workflow.

pipeline:
* add components released during oxford and bump versions
workflow:
* remove the `workflow_dispatch` and the updatecli PR target action
* remove matrix to have a single PR and avoid triggering numerous actions on each commit. Even if we have concurrency only the first commit is built which do not make sense (as we would prefer building the branch in its final state), and we can't cancel job to build only on the last commit other wise we might have leftover releases on acs-cluster.

### DIFF
--- a/.github/workflows/bumpVersions.yml
+++ b/.github/workflows/bumpVersions.yml
@@ -9,30 +9,20 @@ on:
         required: true
   push:
     paths:
-      - '.github/workflows/bumpVersions.yml'
-      - 'updatecli.d/**'
+      - .github/workflows/bumpVersions.yml
+      - updatecli.d/**
     branches:
-      - 'OPSEXP-**'
+      - OPSEXP-**
 
 permissions:
-  contents: "write"
-  pull-requests: "write"
+  contents: write
+  pull-requests: write
 
 jobs:
   updatecli:
     runs-on: ubuntu-latest
-    strategy:
-      max-parallel: 1
-      matrix:
-        version:
-          - community
-          - 7.0.N
-          - 7.1.N
-          - 7.2.N
-          - 7.3.N
-          - latest
     env:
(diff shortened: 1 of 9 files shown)


.github/workflows/bumpVersions.yml
```diff
@@ -4,35 +4,28 @@ name: Bump component versions
 on:
   workflow_dispatch:
```
gionn, line 5: I would think a bit about how we run the updatecli pipeline here, given that:

* we **must** run it once every release, we **should** run it on a schedule
* there is still the yaml reformat issue we should take care manually
* you simplified the apply from the user perspective by moving the loop inside the template

wouldn't be better to follow the same approach we currently have in the ansible repo where we expect the user to run updatecli locally, double check that everything is as expected and then raise the PR?

this is already a giant leap forward from where we started

otherwise LGTM and decide later

README.md
```diff
@@ -101,9 +101,25 @@ Please use [this guide](CONTRIBUTING.md) to make a contribution to the project a
...
+  export QUAY_USERNAME="<VALUE>"
+  export QUAY_PASSWORD="<VALUE>"
+  export UPDATECLI_GITHUB_TOKEN="<VALUE>"
+  export GIT_AUTHOR_EMAIL='build@alfresco.com'
+  export GIT_AUTHOR_USERNAME='alfresco-build'
+  export GIT_BRANCH="$(git branch --show-current)"
+  updatecli apply -c updatecli.d/uber-manifest.tpl -v updatecli.d/supported-matrix.yaml
+  ```
+
+* Restore formatting offiles edited by updatecli (See [updatecli
```
gionn, line 121:
```suggestion
* Restore formatting of files edited by updatecli (See [updatecli
```

gionn (approved):

PR #955, opened 2023-06-27
OPSEXP-2116 acs-deployment to relay on chart alfresco-connector-ms365

slohe1:


helm/alfresco-content-services/values.yaml
```diff
@@ -122,7 +122,9 @@ repository:
 apiexplorer:
   ingress:
     path: /api-explorer
-ooiService:
+# -- Enable/Disable Alfresco Collaboration Connector for Microsoft 365
+alfresco-connector-ms365:
+  enabled: false
```
gionn, line 127: see the ugly helm-docs generated, you want to move the comment under enabled

we also need to bump the doc where we declare upgrades with non backward compatible changes

gionn (changes requested):

helm/alfresco-content-services/values.yaml
```diff
@@ -122,7 +122,9 @@ repository:
 apiexplorer:
   ingress:
     path: /api-explorer
-ooiService:
+alfresco-connector-ms365:
+  enabled: false
+# -- Enable/Disable Alfresco Collaboration Connector for Microsoft 365
```
alxgomz, line 127: doc annotation must be before the node you want to deocument

helm/alfresco-content-services/values.yaml
```diff
@@ -122,7 +122,9 @@ repository:
 apiexplorer:
   ingress:
     path: /api-explorer
-ooiService:
+alfresco-connector-ms365:
+  enabled: false
```
alxgomz, line 126: So you decided to move th toggler under the "configurator" node `alfresco-connector-ms365`?
I like the idea but there may be impact doing that (not sure backward compat will be that easy to ensure)

helm/alfresco-content-services/values.yaml
```diff
@@ -122,7 +122,9 @@ repository:
...
+# -- Enable/Disable Alfresco Collaboration Connector for Microsoft 365
   replicaCount: 2
   nodeSelector: {}
```
alxgomz, line 129: Only specify the configuration that differs from the component chart AND the image & tag (so the updatecli pipelines are always deterministics). Remove everything else

helm/alfresco-content-services/values.yaml
```diff
@@ -160,6 +162,12 @@ ooiService:
     initialDelaySeconds: 10
     periodSeconds: 20
     timeoutSeconds: 10
+  repository:
+    # -- ACS repository host
+    host: acs-alfresco-cs-repository
```
alxgomz, line 167: That will break in case someone wants to use a release name that's not `acs`. You'll probably need to use a `nameOverride`

alxgomz (changes requested):

helm/alfresco-content-services/values.yaml
gionn: and you want to document the `enabled` field otherwise the helm-docs generated with annotation on non-leaf are very ugly

helm/alfresco-content-services/Chart.yaml
```diff
@@ -53,4 +53,8 @@ dependencies:
     version: 1.2.0
     repository: https://alfresco.github.io/alfresco-helm-charts/
     condition: alfresco-search-enterprise.enabled
+  - name: alfresco-connector-ms365
+    version: 0.1.0-alpha.0
```
gionn, line 57: it's probably better to wait for https://github.com/Alfresco/alfresco-helm-charts/pull/62 where I am releasing the updated chart

gionn (changes requested):

alxgomz (commented):
There is no meaningful test here.
We should try that branch on develop branch of the integration pipeline given it's the only one that can really deploy and configure properly the instances.
Doing that require some communication that we will break develop

slohe1:
image

helm/alfresco-content-services/values.yaml
```diff
@@ -777,9 +739,14 @@ alfresco-control-center:
     limits:
       cpu: "1"
       memory: "1024Mi"
-# -- Enable/Disable Alfresco Collaboration Connector for Microsoft 365
-ooi:
+alfresco-connector-ms365:
   enabled: false
+# -- Enable/Disable Alfresco Collaboration Connector for Microsoft 365
```
slohe1, line 744: Comment has been provided here, otherwise helm docs create a bit ugly doc.

alxgomz (changes requested):

alxgomz:
@slohe1 , the screenshot shows a doc has been uploaded but it is unclear whether it can be edited online with Office365. Did you validate that?

slohe1:
image

helm/alfresco-content-services/values.yaml
```diff
@@ -773,9 +735,14 @@ alfresco-control-center:
...
       memory: "1024Mi"
-# -- Enable/Disable Alfresco Collaboration Connector for Microsoft 365
-ooi:
+alfresco-connector-ms365:
+  # -- Enable/Disable Alfresco Collaboration Connector for Microsoft 365
   enabled: false
+  image:
+    tag: 2.0.0
+  repository:
+    nameOverride: acs-alfresco-cs-repository
```
gionn, line 744: this is making things working because the release name we test against is `acs` but it should not be set here, the chart should use a named template which include the release name to generate the final name

gionn (changes requested):

helm/alfresco-content-services/values.yaml
alxgomz: From what I can see in the ms365 chart there is no support for expanding the repo address from a `nameOverride`. The only option we have for now is to pass `repository.host` & `repository.port` as values.

Previous chart was also using that simplistic approach with a static value. As @gionn pointed out, the default value will only work for the default release name `acs`we use in different places. That is lame but was already the case.

Make sure to:
* revert using host & port values
* make it clear in the documentation that in order to use ms365 connector one needs to pass the static name of the service

alxgomz (changes requested):

PR #960, opened 2023-07-04
OPSEXP-1861: add new ats dedicated chart

alxgomz:
Ref: OPSEXP-1861


alxgomz:
Requires: https://github.com/Alfresco/flux-alfresco-pipeline/pull/93

docs/helm/upgrades.md
```diff
@@ -7,6 +7,52 @@ release notes that are available via [GitHub Releases](https://github.com/Alfres
...
+Must be turned into:
+
+```yaml
+alfresco-transform-service:
+  transformmisc:
+    enabled: true
+    ...
+```
+
+and should you want to disable jodconverter transformers you can do so by using:
```
gionn, line 41: ?

helm/alfresco-content-services/7.0.N_values.yaml
```diff
@@ -2,30 +2,38 @@
...
   image:
     tag: 0.16.1
+alfresco-transform-service:
+  transformrouter: *ats-router
+  pdfrenderer: *ats-pdf
+  imagemagick: *ats-im
+  libreoffice: *ats-lo
+  tika: *ats-tika
+  transformmisc: *ats-misc
+  filestore: *ats-fs
```
gionn, line 36: I would put actual values here and move the previous section at the bottom of the file, with a comment which state they are there only for backward compatibility - so we can drop retrocompat easily later and users inspecting these files will not be confused by finding the former structure first

gionn (changes requested):

docs/helm/upgrades.md
alxgomz: What's the problem here?

gionn:
```suggestion
and should you want to disable jobconverter transformers you can do so by using:
```

helm/alfresco-content-services/7.0.N_values.yaml
gionn: you marked as resolved this but don't see any new commits

alxgomz: oops thank to pre-commit push didn't happen.

docs/helm/upgrades.md
alxgomz: no no this is actually joDconverter: https://github.com/jodconverter/jodconverter

updatecli-matrix-targets.yaml
```diff
@@ -13,21 +13,21 @@ matrix:
       compose_key: $.services.alfresco.image
       helm_target: &helmvalues231 >-
         helm/alfresco-content-services/values.yaml
-      helm_key: $.repository.image.tag
+      helm_key: .repository.image.tag
```
gionn, line 16: why? updatecli complains without leading `$`

helm/alfresco-content-services/templates/config-transformer-routes.yaml
```diff
@@ -4,8 +4,6 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{ template "alfresco.shortname" . }}-transform-routes-configmap
-  labels:
-    {{- include "transform-router.labels" . | nindent 4 }}
```
gionn, line 8: why removing this without a replacement? we will forget for sure once we migrate it otherwise

helm/alfresco-content-services/7.3.N_values.yaml
```diff
@@ -32,6 +32,14 @@ share:
...
   image:
     tag:
+alfresco-transform-service:
+  transformrouter: *ats-router
+  pdfrenderer: *ats-pdf
+  imagemagick: *ats-im
+  libreoffice: *ats-lo
+  tika: *ats-tika
+  transformmisc: *ats-misc
+  filestore: *ats-fs
```
gionn, line 42: why not applying the same approach of values.yaml also here?

gionn (changes requested):

helm/alfresco-content-services/7.3.N_values.yaml
alxgomz: I just forgot about it.

helm/alfresco-content-services/templates/config-transformer-routes.yaml
alxgomz: These labels are not used as far as I can tell so I decided to remove them

updatecli-matrix-targets.yaml
alxgomz: I just valiate that this is ue to the target yaml node having an anchor

gionn (approved):

PR #973, opened 2023-07-18
OPSEXP-2188 integrate with newer dedicated share chart

alxgomz:
Ref: OPSEXP-2188
introduce major change: `external*` values => `global.known_urls`

Requires: https://github.com/Alfresco/terraform-alfresco-pipeline/pull/1236


alxgomz:
tested on pipeline
image

helm/alfresco-content-services/Chart.yaml
```diff
@@ -5,7 +5,7 @@
 ---
 apiVersion: v2
 name: alfresco-content-services
-version: 6.1.0-SNAPSHOT
+version: 7.0.0-SNAPSHOT
```
gionn, line 8: 👍🏻

helm/alfresco-content-services/templates/config-repository-infrastructure.yaml
```diff
@@ -0,0 +1,10 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: infrastructure-repository
+  labels:
+    {{- include "repository.labels" . | nindent 4 }}
+data:
+  repo_svc_name: {{ template "content-services.shortname" . }}-repository
+  repo_svc_port: {{ .Values.repository.service.externalPort | quote }}
```
gionn, line 10: maybe it's an overkill, but we could use the values instead of hardcoding keys here

docs/helm/upgrades.md
```diff
@@ -1,14 +1,18 @@
...
 To get an overview of the changes in each release, first take a look at the
 release notes that are available via [GitHub Releases](https://github.com/Alfresco/acs-deployment/releases).
 
-Here follows a more detailed explanation of any breaking change grouped by version in which they have been released.
+Here follows a more detailed explanation of any breaking change grouped by
+version in which they have been released.
 
 ## To be released (likely 7.0.0)
 
+Due to breaking changes the next version  will be a major release
```
gionn, line 14:
```suggestion
Due to breaking changes the next version will be a major release.
```

docs/helm/upgrades.md
```diff
@@ -21,6 +25,79 @@ Accordingly to this chart rename, also the related values has been moved from
...
+
+These were simple strings and did not allow for multiple domains/URLs to be
+used which is exactly one would be expect to do when configuring cross-origin
+protections.
+This is now possible as we introduce the newer `known_urls` value.
+This value can be iether passed to sshare or repo directly or set in the global
+context. Being able to use either the subchart or the global context Allows for
+conveniently set this parameter at once for all subcharts or on the contrary
+use different values different subcharts if they do not trust the same
+domains/URLs.
```
gionn, line 46:
```suggestion
In previous versions of the charts one could use the values bellow to tell the
charts how to configure Alfresco repo & Share, specifically for security
features (CSRF, CORS):

* `externalHost`
* `externalPort`
* `externalProtocol`

These were simple strings and did not allow for multiple domains/URLs to be
used which is exactly one would be expect to do when configuring cross-origin
protections.
This is now possible as we introduce the newer `known_urls` value.
This value can be either passed to share or repo directly or set in the global
context. Being able to use either the subchart or the global context Allows for
conveniently set this parameter at once for all subcharts or on the contrary
use different values different subcharts if they do not trust the same
domains/URLs.
```

docs/helm/upgrades.md
```diff
@@ -21,6 +25,79 @@ Accordingly to this chart rename, also the related values has been moved from
...
+
+global setting:
```
gionn, line 48:
```suggestion
Global setting:
```

docs/helm/upgrades.md
```diff
@@ -21,6 +25,79 @@ Accordingly to this chart rename, also the related values has been moved from
...
+
+```yaml
+global:
+  known_urls:
+    - https://ecm.domain.tld
+    - http://app.domain.local:8080/crm
+```
+
+targeted setting:
```
gionn, line 57:
```suggestion
Targeted setting:
```

docs/helm/upgrades.md
```diff
@@ -21,6 +25,79 @@ Accordingly to this chart rename, also the related values has been moved from
...
+    - http://app.domain.local:8080/crm
+```
+
+**The first item in the list will be considered the main domain where alfresco
+& Share are installed**
+
+If you want to use the `--set` switch of the `helm` command`, well first of all
+you should consider stopping doing that and use a local values override file
+instead, otherwise you can still do it by usig either of the syntax below to
+pass a list:
```
gionn, line 75: > well first of all you should consider stopping doing that and use a local values override file
instead

lol doesn't feel enough polite for user docs -but I would just present below the way to use a `--values` file so readers would straightly do that (do not even mention --set)

docs/helm/upgrades.md
```diff
@@ -21,6 +25,79 @@ Accordingly to this chart rename, also the related values has been moved from
...
+```bash
+helm install alfresco-content-services acs \
+  --set global.known_urls\[0\]=https://ecm.domain.tld
+  --set global.known_urls\[1\]=http://app.domain.local:8080/crm
+```
+
+### Alfresco Share chart
+
+Alfresco is now deployed as a separate chart and can be enabled/disabled. In
+order to do so simply use the value bellow:
```
gionn, line 94:
```suggestion
Alfresco is now deployed as a separate chart and can be enabled/disabled. In
order to do so simply use the value below:
```

gionn (changes requested):

docs/helm/upgrades.md
alxgomz: I knew someone would find it harsh but I would not have think it would be you :) ! I will re-phrase.

As to whether we should advertise `--set-` I think it's important to do it as the syntax is not super intuitive and plenty of our doc doc still rely on it

docs/helm/upgrades.md
alxgomz: Interesting review. I learnt that"bellow" is actually a verb!

helm/alfresco-content-services/templates/config-repository-infrastructure.yaml
alxgomz: what values?

gionn: I mean to not hardcode/duplicate `repo_svc_name` and `repo_svc_port` here since you have values for them

gionn (approved):

helm/alfresco-content-services/templates/config-repository-infrastructure.yaml
alxgomz: Ah ok! TBH I didn't even realized I could do that as this kind of goes against the logic I had in mind which is that the configmap is not particularly linked to the share subchart but would be something one can reuse/enrich  to use in another subchart. Actually it's more linked to the fact the parent chart is using the repo component (eventually another subchart), than to share. So it would not make sense the use Share value more than another.

PR #977, opened 2023-07-27
OPSEXP-2214 Fixup updatecli helm keys

gionn:
Ref: OPSEXP-2214

Not solving the issue but at least config is consistent


alxgomz (commented):
Can you create/reference the task to address that and finally have a working updatecli pipeline again?

gionn:
> Can you create/reference the task to address that and finally have a working updatecli pipeline again?

it's already referenced, just the branch name is still pointing to the original issue where I discovered it

PR #982, opened 2023-08-02
Update eksctl command usage in eks-deployment.md

gionn:


gionn:
Fixed in https://github.com/Alfresco/acs-deployment/pull/1138

PR #1012, opened 2023-09-20
OPSEXP-2311 Provide an input to trigger alternative updatecli ref

gionn:
Ref: OPSEXP-2311

Just to run more easily the workflow against a custom alfresco-updatecli branch

### DIFF
--- a/.github/workflows/bumpVersions.yml
+++ b/.github/workflows/bumpVersions.yml
@@ -9,6 +9,11 @@ on:
     tags-ignore:
       - '**'
   workflow_dispatch:
+    inputs:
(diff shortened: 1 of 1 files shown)


helm/alfresco-content-services/values.yaml
```diff
@@ -251,7 +251,7 @@ alfresco-search:
   enabled: true
   searchServicesImage:
     repository: quay.io/alfresco/search-services
-    tag: 2.0.8
+    tag: 2.1.0-A1
```
gionn, line 254: was released by mistake, I deleted the quay tags

PR #1016, opened 2023-09-21
OPSEXP-2242 move search values to global context and add support for external

alxgomz:
Ref: OPSEXP-2242

reintroducing external search suport


gionn (commented):
check https://github.com/Alfresco/acs-deployment/actions/runs/6260123554/job/16997564044?pr=1016#step:13:44

gionn (approved):

PR #1022, opened 2023-09-26
OPSEXP-2242: add outgoing mail config for repo

alxgomz:
OPSEXP-2242

### DIFF
--- a/helm/alfresco-content-services/templates/config-repository.yaml
+++ b/helm/alfresco-content-services/templates/config-repository.yaml
@@ -29,6 +29,19 @@ data:
     {{- if index .Values "alfresco-transform-service" "enabled" }}
     {{- include "alfresco-content-services.atsConfig" . | indent 4 }}
     {{- end }}
+    {{- with .Values.global.mail }}
+    {{- if or .host }}
+    mail.host={{ .host }}
+    mail.port={{ .port }}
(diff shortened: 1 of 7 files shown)


helm/alfresco-content-services/templates/config-repository.yaml
```diff
@@ -29,6 +29,19 @@ data:
     {{- if index .Values "alfresco-transform-service" "enabled" }}
     {{- include "alfresco-content-services.atsConfig" . | indent 4 }}
     {{- end }}
+    {{- with .Values.global.mail }}
+    {{- if or .host }}
```
gionn, line 33: or is not necessary

helm/alfresco-content-services/values.yaml
```diff
@@ -115,6 +138,9 @@ alfresco-repository:
     # -- Specify a storageClass for dynamic provisioning
     accessModes:
       - ReadWriteMany
+  environment:
+    JAVA_OPTS: >-
+      -Dmail.password=$MAIL_PASSWORD
```
gionn, line 143: not sure I like it, because if the user need to provide their own JAVA_OPTS needs to keep the default ones

if we need it, can we keep it where it was? (or somewhere else but not in the values like if is worth changing it from the user)

gionn (approved):

PR #1025, opened 2023-10-06
OPSEXP-2242: charts dependencies bump and doc update

alxgomz:


gionn (approved):
pushed a commit with some more docs fixes

PR #1047, opened 2023-11-20
OPSEXP-2403 Bump tags to fix activemq CVE-2023-46604

gionn:
Ref: OPSEXP-2403

PR #1048, opened 2023-11-21
OPSEXP-2403 Fix community workflow not really using community values

gionn:
Ref: OPSEXP-2403

Exposed from a failure in https://github.com/Alfresco/acs-deployment/pull/1047

### DIFF
--- a/helm/alfresco-content-services/community_values.yaml
+++ b/helm/alfresco-content-services/community_values.yaml
@@ -1,7 +1,6 @@
 # This values file can be used to install the latest ACS Community version using
 # the latest version of the chart
 alfresco-repository:
-  edition: Community
   replicaCount: 1
   image:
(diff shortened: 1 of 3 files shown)


helm/alfresco-content-services/community_values.yaml
```diff
@@ -1,6 +1,7 @@
 # This values file can be used to install the latest ACS Community version using
 # the latest version of the chart
 alfresco-repository:
+  edition: Community
```
alxgomz, line 4: AFAIR this is not used anymore and can be removed (was only used in the `NOTES.txt`).

gionn: indeed there are no references in the charts, removing it

gionn, line 4:
```suggestion
```

PR #1050, opened 2023-11-21
OPSEXP-2403 Release helm charts v7.0.1

gionn:


helm/alfresco-content-services/Chart.yaml
```diff
@@ -4,7 +4,7 @@
 # https://www.alfresco.com/platform/content-services-ecm/trial/download
 apiVersion: v2
 name: alfresco-content-services
-version: 7.0.0
+version: 7.0.1
 appVersion: 23.1.0
```
alxgomz, line 8: change appVersion

```suggestion
appVersion: 23.1.1
```

alxgomz (changes requested):

helm/alfresco-content-services/Chart.yaml
gionn: it's in the other PR, that's why this one is still in draft and the other not

PR #1068, opened 2024-01-11
OPSEXP-2300: bump alfresco-sync-service chart to 5.0.0-alpha.1

alxgomz:
Ref: OPSEXP-2300

### DIFF
--- a/helm/alfresco-content-services/Chart.lock
+++ b/helm/alfresco-content-services/Chart.lock
@@ -25,7 +25,7 @@ dependencies:
   version: 2.0.0
 - name: alfresco-sync-service
   repository: https://alfresco.github.io/alfresco-helm-charts/
-  version: 4.5.1
+  version: 5.0.0-alpha.0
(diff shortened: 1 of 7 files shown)


gionn (changes requested):
sorry just merged mine, you need to regenerate the lock

helm/alfresco-content-services/Chart.lock
```diff
@@ -5,6 +5,9 @@ dependencies:
 - name: postgresql
   repository: oci://registry-1.docker.io/bitnamicharts
   version: 12.8.5
+- name: postgresql
+  repository: oci://registry-1.docker.io/bitnamicharts
+  version: 12.8.5
```
gionn, line 10: I wonder how helm could work with that

alxgomz: From what I see it make Helm download the same chart twice which is indeed sub-optimal but this file is autogenerated so...

gionn (approved):

PR #1071, opened 2024-01-12
OPSEXP-2291 Bump alfresco-search to 3.0.0-alpha.0

gionn:
Ref: OPSEXP-2291

### DIFF
--- a/helm/alfresco-content-services/templates/secret-search.yaml
+++ b/helm/alfresco-content-services/templates/secret-search.yaml
@@ -9,9 +9,7 @@ metadata:
     {{- include "alfresco-content-services.labels" $ | nindent 4 }}
 type: Opaque
 data:
-{{- if eq "secret" .securecomms }}
   SOLR_SECRET: {{ .sharedSecret | default "" | b64enc | quote }}
-{{- end }}
 {{- if eq "elasticsearch" (include "alfresco-content-services.search.flavor" $) }}
   SEARCH_USERNAME: {{ .username | default "" | b64enc | quote }}
   SEARCH_PASSWORD: {{ .password | default "" | b64enc | quote }}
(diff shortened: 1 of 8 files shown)


helm/alfresco-content-services/templates/secret-search.yaml
```diff
@@ -9,9 +9,7 @@ metadata:
     {{- include "alfresco-content-services.labels" $ | nindent 4 }}
 type: Opaque
 data:
-{{- if eq "secret" .securecomms }}
   SOLR_SECRET: {{ .sharedSecret | default "" | b64enc | quote }}
-{{- end }}
```
alxgomz, line 14: I don't understand why you want to remove this one

gionn: is the counterpart of removing the [same check in the deployment](https://github.com/Alfresco/alfresco-helm-charts/pull/179/files#diff-ad152f17cbaac4f906353bbab381f139d15b1b5f9f599cc23863cd9c5e4565e9L54), otherwise it would fail for missing key when securecomms is none.

we can't really have conditionals on this since users could provide it also via configmap, which can't be used in conditionals and would force them to provide the same setting also in values (had a failure in the CI on older acs versions)

alxgomz: Then I think we should instead make the env optional in the deployment Otherwise we force people who want to use the component chart to set SOLR vars they don't need to empty values

gionn: optional secret make sense https://github.com/Alfresco/alfresco-helm-charts/pull/192

here I think it's reasonable to revert assuming that if user provide a custom configmap is also going to provide a custom secret (when needed)

alxgomz (approved):

PR #1074, opened 2024-01-18
OPSEXP-2440: fix Alfresco Intelligence Service

alxgomz:
Ref: OPSEXP-2440


docs/helm/examples/with-ai.md
```diff
@@ -22,18 +22,52 @@ by AIS.
 
 ## Deploy
 
+Create a local values file to contain Helm charts' configuration options (`ai-values.yaml`):
+
+```yaml
+alfresco-ai-transformer:
+  enabled: true
+  fullnameOverride: alfresco-intelligence-service
```
gionn, line 30:
```suggestion
```

it's already in default values

helm/alfresco-content-services/values.yaml
```diff
@@ -518,6 +518,8 @@ alfresco-ai-transformer:
   # -- toggle deploying Alfresco ai transformer for more details about configuration check
   # https://github.com/Alfresco/alfresco-helm-charts/tree/main/charts/alfresco-ai-transformer
   enabled: false
+  # -- Enforce static resource naming in AIS so the ATS trouter can be given the URL of the service
+  fullnameOverride: alfresco-intelligence-service
```
gionn, line 522: not sure why we need this when elsewhere there is plenty of components that needs to know other components URL - just reuse the named template which generate the correct/same service name?

gionn (changes requested):

helm/alfresco-content-services/values.yaml
alxgomz: I used the `fullnameOverride` because the service name needs is passed as a generic  `.Values.environment.JAVA_OPTS` in the ATS chart. so I can't use templates, unless we change ATS chart to evaluate `.Values.environment.JAVA_OPTS` with tpl()

docs/helm/examples/with-ai.md
```diff
@@ -22,18 +22,51 @@ by AIS.
...
+alfresco-transform-service:
+  filestore:
+    persistence.enabled: true
+    storageClass: nfs-client
+  transformrouter:
+    environment: |-
+      JAVA_OPTS: -XX:MaxRAMPercentage=80
+      TRANSFORMER_URL_AWS_AI: http://alfresco-intelligence-service
+      TRANSFORMER_QUEUE_AWS_AI: "org.alfresco.transform.engine.ai-aws.acs"
+      TRANSFORMER_ROUTES_ADDITIONAL_AI: "/etc/ats-routes/ai-pipeline-routes.json"
```
gionn, line 45: these should not be a string

gionn (changes requested):
can you check the current usage of the old .global.ai.enabled we should be able to get rid of it

gionn:
> This used to be necessary for the common chart (the ADW one) to enable AI features

we can just replace it with the .enabled of the ai chart?

alxgomz:
> > This used to be necessary for the common chart (the ADW one) to enable AI features
> 
> we can just replace it with the .enabled of the ai chart?

No because it's evaluated in the context of the common chart whic can't access the `.Values.alfresco-ai-transformer.enabled` value

gionn (dismissed):

PR #1076, opened 2024-01-22
OPSEXP-2445 Fix flaky docker compose enterprise tests

gionn:
* Add docker compose enterprise timeout and containers debug
* Add -XX:MaxRAM workaround [ACS containers and cgroup v2 in ACS up to 7.2](https://hub.alfresco.com/t5/alfresco-content-services-blog/acs-containers-and-cgroup-v2-in-acs-up-to-7-2/ba-p/318039)

Ref: OPSEXP-2445
Requires: https://github.com/Alfresco/alfresco-build-tools/pull/478

### DIFF
--- a/.github/workflows/docker-compose-enterprise.yml
+++ b/.github/workflows/docker-compose-enterprise.yml
@@ -56,7 +56,7 @@ jobs:
           quay_username: ${{ secrets.QUAY_USERNAME }}
           quay_password: ${{ secrets.QUAY_PASSWORD }}
       - name: Debug alfresco on failure
-        if: failure() || cancelled()
+        if: always()
         working-directory: docker-compose
         run: |
           docker-compose -f ${{ matrix.compose_file }} logs alfresco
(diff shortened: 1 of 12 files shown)


.github/workflows/docker-compose-enterprise.yml
```diff
@@ -48,9 +52,29 @@ jobs:
...
+        if: always()
+        working-directory: docker-compose
+        run: |
+          docker-compose -f ${{ matrix.compose_file }} logs alfresco
+
+      - name: Show solr logs
+        if: always()
+        working-directory: docker-compose
+        run: |
+          docker-compose -f ${{ matrix.compose_file }} logs solr6
```
alxgomz, line 80: Why specificly those?
We could get all logs and upload a artifacts instead?

alxgomz (changes requested):

.github/workflows/docker-compose-enterprise.yml
gionn: indeed but I guess at this point that the issue has been solved are not really worth keeping it

I would like to improve the dbp-charts/verify-compose first which ATM is just a monolith bash script, could be a fit for compose improvements epic

alxgomz (approved):

PR #1080, opened 2024-01-23
OPSEXP-2439 Helm test with dtas

gionn:
Ref: OPSEXP-2439


test/enterprise-integration-test-values.yaml
```diff
@@ -68,7 +68,7 @@ alfresco-transform-service:
         memory: "512Mi"
       limits:
         cpu: "2"
-        memory: "1024Mi"
+        memory: "2048Mi"
```
gionn, line 71:
```suggestion
        memory: "1024Mi"
```

test/enterprise-integration-test-values.yaml
```diff
@@ -79,7 +79,7 @@ alfresco-transform-service:
         memory: "256Mi"
       limits:
         cpu: "2"
-        memory: "512Mi"
+        memory: "1Gi"
```
gionn, line 82:
```suggestion
        memory: "512Mi"
```

test/enterprise-integration-test-values.yaml
```diff
@@ -127,9 +127,16 @@ alfresco-sync-service:
...
     secretAccessKey: ZZZZZZZZZZZZZZZZZ
     s3Bucket: somebucket
     region: us-east-1
     comprehendRoleARN: arn:aws:iam::000000000000:user/comprehend
+dtas:
+  enabled: true
+  additionalArgs:
+    # https://alfresco.atlassian.net/browse/ACS-6670
+    - -k not test_renditions
+    - --tb=auto
```
gionn, line 142:
```suggestion
```

helm/alfresco-content-services/23.N_values.yaml
```diff
@@ -75,3 +75,8 @@ alfresco-connector-ms365:
 alfresco-connector-msteams:
   image:
     tag: 2.0.0
+dtas:
+  config:
+    assertions:
+      acs:
+        version: 23.1.1
```
alxgomz, line 82: Shouldn't we rely or at least default to `.Chart.appVersion`?

alxgomz (commented):
I'm not a big fan of having test related info in values file.
I'd rather we have these in separate value files under the test folder

helm/alfresco-content-services/23.N_values.yaml
gionn: for this stuff I was thinking to just add support for updatecli later

gionn:
> I'd rather we have these in separate value files under the test folder

I was already thinking that eventually we'll make dtas public and could really be run by default upon installs.

On another side note, I am not really sure that explicitly checking installed versions as part of DTAS really bring some value, the main objective should be to verify the everything is behaving correctly, and not that a version I've written in yaml file is the same I've written in another yaml file 😀

helm/alfresco-content-services/23.N_values.yaml
alxgomz: well relying on appVersion would avoid having having to create and maintain more updatecli pipelines.

gionn: I can add it as default in the template but I would keep the value as eventually it would need to be overridden for -A.x/-M.x versions

alxgomz: ok no worries, then let's move on as is

.github/workflows/helm-enterprise.yml
```diff
@@ -131,7 +145,8 @@ jobs:
           kubectl get all --all-namespaces
           kubectl describe pod
 
-      - uses: nick-fields/retry@v2
+      - name: Run Newman tests
+        uses: nick-fields/retry@v2
```
alxgomz, line 149: full sha M. security champ :) pls

helm/alfresco-content-services/values.yaml
```diff
@@ -534,3 +534,31 @@ elasticsearch:
   enabled: false
   replicas: 1
   clusterHealthCheckParams: "wait_for_status=yellow&timeout=1s"
+
+dtas:
+  # -- Enables the deployment test suite which can run via `helm test` (currently available for Enterprise only)
+  enabled: false
+  image:
+    repository: quay.io/alfresco/alfresco-deployment-test-automation-scripts
+    tag: master
```
alxgomz, line 543: we should tag something here

alxgomz (changes requested):

alxgomz (approved):

PR #1091, opened 2024-02-13
OPSEXP-2391 Document how to provision with an external keycloak

gionn:
Ref: OPSEXP-2391


alxgomz (approved):
I'm not a big fan of introducing that new config key but tbh I have no better idea

PR #1095, opened 2024-02-14
OPSEXP-2449 Deploy search enterprise by default

gionn:
* Previous versions and community still deploy solr
* Validate repository search flavor to avoid inconsistencies
* Fix wrong secret passed as messageBroker secret

Ref: OPSEXP-2449


helm/alfresco-content-services/templates/_helpers-search.tpl
alxgomz, line 1: IF I want to dpeloy acs with solr & ES and use solr i nthe first place (e.g. I'm preparing a migration scenario) then that is not possible anymore with this change I think.
Are you trying to fix a bug or just had more consistency (less flexibility)?

docs/helm/upgrades.md
```diff
@@ -10,6 +10,11 @@ Releases](https://github.com/Alfresco/acs-deployment/releases).
 Here follows a more detailed explanation of any breaking change grouped by
 version in which they have been released.
 
+## Unreleased
+
+* Search Enterprise is now the default search engine when installing Enterprise
```
alxgomz, line 15: Can you add a small doc on how to configure an external ES index and link to it from here?

alxgomz (changes requested):

helm/alfresco-content-services/templates/_helpers-search.tpl
gionn: if that's a scenario we want to support then I can relax the check when both search and search enterprise are enabled.

I want to avoid the situation where repository is configured for a search backend which is not the currently enabled based on `.enabled` values, which is causing a failure only at deployment stage where repository container cannot be created due to missing keys in the configmap.

pmacius (approved):

alxgomz (approved):

PR #1104, opened 2024-02-27
OPSEXP-2462 Create new values file for pre-released versions and make default values file only for released images

pmacius:
!Ref OPSEXP-2462


pmacius:
Please take a look if this looks all right, then I will test pipeline bumping versions :)

alxgomz (changes requested):
version based values & docker files should not contain pre-release versions. Was the updatecli pipelines hanged on purpose while we wait for releases to happen?

docker-compose/pre-release-docker-compose.yml
```diff
@@ -18,7 +18,7 @@
 version: "2"
 services:
   alfresco:
-    image: quay.io/alfresco/alfresco-content-repository:23.1.1
+    image: quay.io/alfresco/alfresco-content-repository:23.2.0-M1
```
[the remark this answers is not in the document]
alxgomz: Argh sorry my bad! I missed the file was renamed

pmacius:
> Was the updatecli pipelines hanged on purpose while we wait for releases to happen?
Yes, I think that it will cause less trouble if we stick to those tags for now and wait for new release.

helm/alfresco-content-services/values.yaml
gionn, line 1: would expect tag changes in this file (from -A to GA versions)

gionn (changes requested):

pmacius:
https://github.com/Alfresco/acs-deployment/actions/runs/8082979440/job/22084993920

updatecli-matrix-targets.yaml
```diff
@@ -9,13 +9,13 @@ matrix:
...
-        docker-compose/docker-compose.yml
+        docker-compose/pre-release-docker-compose.yml
       compose_key: $.services.alfresco.image
       helm_target: &helmvaluesNext >-
-        helm/alfresco-content-services/values.yaml
+        helm/alfresco-content-services/pre-release_values.yaml
       helm_key: $.alfresco-repository.image.tag
       # keep the unused anchor to detect multiple usages when copying blocks for new versions
-      helm_update_appVersion: &helm_update_appVersion true
+      helm_update_appVersion: &helm_update_appVersion false
```
gionn, line 18: remove it, default is false

put this block to `current` (I can explain better why this is needed in private xD)
```
      # keep the unused anchor to detect multiple usages when copying blocks for new versions
      helm_update_appVersion: &helm_update_appVersion true
```

gionn (changes requested):

gionn (approved):

alxgomz (approved):

PR #1119, opened 2024-03-18
OPSEXP-2526: bump alfresco-repository to allow repo hpa and add tests

alxgomz:
Ref: OPSEXP-2526

### DIFF
--- a/.github/workflows/helm-enterprise.yml
+++ b/.github/workflows/helm-enterprise.yml
@@ -125,8 +125,6 @@ jobs:
           fi
 
       - name: Helm install
-        env:
-          VALUES_FILE: ${{ matrix.values }}
         run: |
           helm dep up ./helm/alfresco-content-services &&
           helm install acs ./helm/alfresco-content-services \
@@ -137,7 +135,12 @@ jobs:
             --values test/enterprise-integration-test-values.yaml \
             ${{ steps.cgroupv2-workaround-extra-values.outputs.helm_install_params }} \
             ${{ steps.configurable-extra-values.outputs.helm_install_params }}
-          kubectl get pods --watch-only &
+
+      - name: Watch Helm deployment
+        env:
+          VALUES_FILE: ${{ matrix.values }}
+        run: |
+          kubectl get pods --watch &
           KWPID=$!
           kubectl wait --timeout=7m --all=true --for=condition=Available deploy && kill $KWPID
           if [ "${VALUES_FILE:0:2}" != "7." ] ; then echo "Waiting for ESC Reindexing job to complete"
--- a/.github/workflows/helm-enterprise.yml
+++ b/.github/workflows/helm-enterprise.yml
@@ -113,11 +113,10 @@ jobs:
         id: configurable-extra-values
         run: |
           if [ "${{ matrix.values }}" = "values.yaml" -o "${{ matrix.values }}" = "pre-release_values.yaml" ]; then
-            echo "Enabling clustered tests"
+            echo "Enabling clustered tests via auto-scalling: 2 replicas max, 20s stabilization window"
             echo "helm_install_params=--set alfresco-repository.autoscaling.enabled=true \
               --set alfresco-repository.autoscaling.maxReplicas=2 \
-              --set alfresco-repository.autoscaling.scaleUp.stabilizationWindowSeconds=20 \
-              --set dtas.additionalArgs[0]='-k not test_renditions'" >> "$GITHUB_OUTPUT"
+              --set alfresco-repository.autoscaling.behavior.scaleUp.stabilizationWindowSeconds=20" >> "$GITHUB_OUTPUT"
           fi
           if [ "${{ matrix.values }}" = "7.0.N_values.yaml" ]; then
             echo "Skipping renditions tests on ACS 7.0.N"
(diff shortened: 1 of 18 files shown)


helm/alfresco-content-services/templates/config-repository.yaml
```diff
@@ -8,7 +8,6 @@ data:
   alfresco-global.properties: |
     {{- $alfUrl := include "alfresco-common.external.url" . }}
     deployment.method=HELM_CHART
-    alfresco.cluster.enabled={{ gt (index .Values "alfresco-repository" "replicaCount" | int) 1 }}
```
gionn, line 11: where is this gone?

test/enterprise-integration-test-values.yaml
```diff
@@ -150,23 +150,10 @@ alfresco-sync-service:
...
-  aws:
-    accessKeyId: AAAAAAAAAAAAAAAAA
-    secretAccessKey: ZZZZZZZZZZZZZZZZZ
-    s3Bucket: somebucket
-    region: us-east-1
-    comprehendRoleARN: arn:aws:iam::000000000000:user/comprehend
-  resources:
-    requests:
-      cpu: "0.01"
-      memory: "512Mi"
```
gionn, line 165: in this way we are not testing it anymore in acs-deployment

gionn (changes requested):

.github/workflows/helm-enterprise.yml
```diff
@@ -98,49 +99,60 @@ jobs:
...
       - name: Check if cgroup v2 workaround is needed
+        if: startsWith(matrix.values, '7.')
+        env:
+          VALUES_FILE: ${{ matrix.values }}
         id: cgroupv2-workaround-extra-values
         run: |
-          acs_version=$(echo ${{ matrix.values }} | cut -d'.' -f1,2)
-          is_affected=$(echo "$acs_version <= 7.2" | bc -l)
-          if [ "$is_affected" -eq 1 ]; then
+          if [ ${VALUES_FILE:2:1} -le 2 ]; then
```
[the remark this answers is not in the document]
alxgomz: the goal was mostly to get rid of the warnings for non versioned value files:
https://github.com/Alfresco/acs-deployment/actions/runs/8282284667/job/22662768832#step:9:11

test/enterprise-integration-test-values.yaml
alxgomz: True but this test do not add any value compared to the component chart's tests we run with `ct`

.github/workflows/helm-enterprise.yml
```diff
@@ -98,49 +99,60 @@ jobs:
...
+            --values helm/${{ matrix.name }}/${{ matrix.values }} \
+            --values test/enterprise-integration-test-values.yaml \
+            ${{ steps.cgroupv2-workaround-extra-values.outputs.helm_install_params }} \
+            ${{ steps.configurable-extra-values.outputs.helm_install_params }}
+          kubectl get pods --watch-only &
+          KWPID=$!
+          kubectl wait --timeout=7m --all=true --for=condition=Available deploy && kill $KWPID
+          if [ "${VALUES_FILE:0:2}" != "7." ] ; then echo "Waiting for ESC Reindexing job to complete"
+            kubectl wait --timeout=5m --for=condition=complete job/acs-alfresco-search-enterprise-reindexing
+          fi
```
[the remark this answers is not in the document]
alxgomz: I can't really do that.
The goal here is to have logs that sequentially showing the pod status changes one after the other. This prove useful when I tried to debug overly long installations. If I restore the previous command the steps will silently install  up to completion and only after wards we spit a status that much harder to understand when it comes to finding which component took more time to spin up.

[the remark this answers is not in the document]
gionn: I meant reverting just the formatting, I think it's still worth having a step for the helm install (without wait) and another one with the stuff you introduced, along with a step name which describe what we are trying to achieve there

test/enterprise-integration-test-values.yaml
alxgomz: I guess I removed it in an attempt to speed up deployment (before I enabled the sequential logging during installation). If you're really sure we can catch something here I can try to re-enable it (though I'd like to know what we can catch here tbh)

gionn: e.g. if we are injecting configmaps in a consistent manner (with the appropriate keys) - which are managed in this chart only

.github/workflows/helm-enterprise.yml
```diff
@@ -98,49 +99,63 @@ jobs:
...
+              --set dtas.additionalArgs[0]='-k not test_renditions'" >> "$GITHUB_OUTPUT"
           fi
           if [ "${{ matrix.values }}" = "7.0.N_values.yaml" ]; then
             echo "Skipping renditions tests on ACS 7.0.N"
             echo "helm_install_params=--set dtas.additionalArgs[0]='-k not test_renditions'" >> "$GITHUB_OUTPUT"
           fi
 
       - name: Helm install
-        run: >-
+        run: |
```
gionn, line 128: revert 🙏🏻

helm/alfresco-content-services/Chart.yaml
gionn, line 1: should we move also the main chart version to 8.1.0-alpha.0?

.github/workflows/helm-enterprise.yml
```diff
@@ -123,24 +128,35 @@ jobs:
         run: >-
           helm dep up ./helm/alfresco-content-services &&
```
gionn, line 129: you can move this to a dedicated step, but more importantly I see only now that it should be `helm dep build` and not `up` 😱  check also for other usages

alxgomz: will be tackled in OPSEXP-2590

gionn (approved):

PR #1130, opened 2024-04-18
OPSEXP-2629 Bump charts and tags for Fornax release

gionn:
Ref: OPSEXP-2629

requires https://github.com/Alfresco/alfresco-updatecli/pull/34

### DIFF
--- a/.github/workflows/helm-enterprise.yml
+++ b/.github/workflows/helm-enterprise.yml
@@ -156,6 +156,7 @@ jobs:
           helm ls --all-namespaces --all
           helm status acs --show-resources
           kubectl describe pod
+          kubectl logs -l app.kubernetes.io/component=alfresco-sync-service --tail=-1
 
       - name: Run Newman tests
         uses: nick-fields/retry@7152eba30c6575329ac0576536151aca5a72780e # v3.0.0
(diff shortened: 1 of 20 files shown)


.github/workflows/helm-enterprise.yml
```diff
@@ -156,6 +156,7 @@ jobs:
           helm ls --all-namespaces --all
           helm status acs --show-resources
           kubectl describe pod
+          kubectl logs -l app.kubernetes.io/component=alfresco-sync-service --tail=-1
```
gionn, line 159:
```suggestion
```

helm/alfresco-content-services/pre-release_values.yaml
```diff
@@ -2,44 +2,44 @@
 # the chart
 alfresco-repository:
   image:
-    tag: 23.2.1
+    tag: 23.3.0-A9
```
pmacius, line 5:
```suggestion
    tag: 23.3.0-A10
```
Looked at pre released docker compose and spotted difference for this image

gionn: we need to stick to A9 due to an issue on A10 that needs to be investigated further

but thanks to this I realized that:
* we were not testing pre-release compose
* there was an issue preventing sync service update in pre-release compose
 
😮‍💨

alxgomz (approved):

PR #1133, opened 2024-04-22
OPSEXP-2633 Update download trial branch to provide ACS 8.1.0

pmacius:
Ref: OPSEXP-2633


gionn (approved):

docker-compose/docker-compose.yml
```diff
@@ -214,7 +214,7 @@ services:
     ports:
       - "8080:8080"
   sync-service:
-    image: quay.io/alfresco/service-sync:4.0.1
+    image: quay.io/alfresco/service-sync:5.0.0
```
alxgomz, line 217: Let me clarify whether that's OK because the repo imaged shipped (23.2.1) has a sync service AMP 4.x.
@kavitshah-gl @gl-lovekesh @abhishekkumar31 Is there a compatibility matrix of sync server / repo AMP that we need to stick to (sorry for "mass mentioning" but I could not find an updated confluence page about the "device support" team structure)?

gionn: https://github.com/Alfresco/docs-alfresco/pull/1453

alxgomz: Right but here we would be providing kind of an hybrid where sync server component is 5.x, repo is 23.2 but AMP is still 4.x.

alxgomz:
@gl-lovekesh confirmed the AMP and the sync service should share the same version and that's the only setup tested. Please raise another PR to revert to 4.0.1

PR #1153, opened 2024-05-27
OPSEXP-2677 Bump tags for Gemini release

pmacius:
Ref: OPSEXP-2677

### DIFF
--- a/helm/alfresco-content-services/7.3.N_values.yaml
+++ b/helm/alfresco-content-services/7.3.N_values.yaml
@@ -89,10 +89,10 @@ dtas:
   config:
     assertions:
       acs:
-        version: 7.3.1
+        version: 7.3.2
         modules:
         - id: org.alfresco.integrations.google.docs
-          version: 3.3.0
+          version: 3.3.1
           installed: true
         - id: alfresco-aos-module
           version: 1.5.0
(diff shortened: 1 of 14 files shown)


helm/alfresco-content-services/7.3.N_values.yaml
```diff
@@ -89,10 +89,10 @@ dtas:
   config:
     assertions:
       acs:
-        version: 7.3.1
+        version: 7.3.2
         modules:
         - id: org.alfresco.integrations.google.docs
-          version: 3.3.0
+          version: 3.4.2
```
gionn, line 95: not sure about this minor bump

gionn (changes requested):
need to bump component chart which needs to include the ats updates

helm/alfresco-content-services/7.3.N_values.yaml
pmacius: Dtas forced it in assertions https://github.com/Alfresco/acs-deployment/actions/runs/9270239299/job/25502784563#step:17:35

gionn (changes requested):
merge also https://github.com/Alfresco/acs-deployment/compare/master...updatecli-bump-helm into here and we should be good

PR #1167, opened 2024-06-18
OPSEXP-2115: bump charts to add consistent default affinity rules

alxgomz:
Ref: OPSEXP-2115


helm/alfresco-content-services/Chart.yaml
```diff
@@ -4,7 +4,7 @@
 # https://www.alfresco.com/platform/content-services-ecm/trial/download
 apiVersion: v2
 name: alfresco-content-services
-version: 8.2.0
+version: 8.3.0
```
pmacius, line 7: Maybe we can release alpha here?

gionn (approved):

PR #1181, opened 2024-08-08
OPSEXP-2766 Increase liveness probes for T-engines to prevent restarting

pmacius:


test/enterprise-integration-test-values.yaml
```diff
@@ -176,5 +176,9 @@ alfresco-ai-transformer:
     requests:
       cpu: "0.01"
       memory: "512Mi"
+  readinessProbe:
+    initialDelaySeconds: 120
```
gionn, line 180: why increasing also the readiness?

gionn (approved):

PR #1186, opened 2024-08-20
OPSEXP-2504 Change default strategy for filestore deployment

gionn:
OPSEXP-2504

### DIFF
--- a/helm/alfresco-content-services/values.yaml
+++ b/helm/alfresco-content-services/values.yaml
@@ -264,6 +264,7 @@ alfresco-transform-service:
     # -- Declares the alfresco-shared-file-store used by the content repository
     # and transform service
     enabled: true
+    # -- Cannot use more than 1 replica when persistence is enabled
     replicaCount: 1
     image:
       repository: quay.io/alfresco/alfresco-shared-file-store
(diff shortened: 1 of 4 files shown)


helm/alfresco-content-services/values.yaml
```diff
@@ -274,6 +275,9 @@ alfresco-transform-service:
       data:
         mountPath: "/tmp/Alfresco"
         subPath: "alfresco-content-services/filestore-data"
+    strategy:
+      # -- Strategy must be set to Recreate when persistence is enabled
+      type: Recreate
```
gionn: not sure this is the best place where to set this given that it really depends on the `accessMode` of the `storageClass`

helm/alfresco-content-services/values.yaml
```diff
@@ -274,6 +275,9 @@ alfresco-transform-service:
       data:
         mountPath: "/tmp/Alfresco"
         subPath: "alfresco-content-services/filestore-data"
+    strategy:
+      # -- Strategy must be set to Recreate when persistence use a `ReadWriteOnce` storageClass
+      type: Recreate
```
gionn, line 280: not sure this is the best place where to set this given that it really depends on the `accessMode` of the `storageClass`

pmacius: I looked up postgres and redis charts they use the same approach, strategy is defined outside of the `storageClass`. Still not sure if this the best practice but looks good

pmacius (approved):

helm/alfresco-content-services/values.yaml
alxgomz: I would have set it her eas well. Though I would have added helm-docs annotations with the opposite thinking: mentioning what to do to set more than one replica instead of why it is set to one.

alxgomz (approved):

PR #1189, opened 2024-08-27
OPSEXP-2778 Bump tags and charts for Indus release

pmacius:
Ref: OPSEXP-2778


gionn (changes requested):
wait for frontend GA - everything else looks ok

gionn (approved):

PR #1194, opened 2024-09-23
OPSEXP-2411: switch docker compose proxy to traefik

alxgomz:
Ref: OPSEXP-2411

### DIFF
--- a/docker-compose/7.1.N-docker-compose.yml
+++ b/docker-compose/7.1.N-docker-compose.yml
@@ -61,11 +61,11 @@ services:
         -XX:MaxRAM=1900m
     labels:
       - "traefik.enable=true"
-      - "traefik.http.routers.alfresco.rule=PathPrefix(`/alfresco`)"
+      - "traefik.http.routers.alfresco.rule=PathPrefix(`/`)"
       - "traefik.http.services.alfresco.loadbalancer.server.port=8080"
-      - "traefik.http.routers.solrapideny.rule=PathRegexp(`^/alfresco/(wc)?s(ervice)?/api/solr/.*$`)"
+      - "traefik.http.routers.solrapideny.rule=PathRegexp(`^/alfresco/(wc)?s(ervice)?/api/solr/`)"
       - "traefik.http.routers.solrapideny.middlewares=fakeauth@docker"
-      - "traefik.http.routers.alfrescomicrometer.rule=PathRegex(`^/alfresco/(wc)?s(ervice)?/prometheus`)"
+      - "traefik.http.routers.alfrescomicrometer.rule=PathRegexp(`^/alfresco/(wc)?s(ervice)?/prometheus`)"
       - "traefik.http.routers.alfrescomicrometer.middlewares=prometheusipfilter@docker"
   transform-router:
     mem_limit: 512m
@@ -132,7 +132,7 @@ services:
       - "traefik.enable=true"
       - "traefik.http.routers.share.rule=PathPrefix(`/share`)"
       - "traefik.http.services.share.loadbalancer.server.port=8080"
-      - "traefik.http.routers.proxiedsolrapideny.rule=PathRegexp(`^/share/proxy/alfresco/(wc)?s(ervice)?/api/solr/.*$`)"
+      - "traefik.http.routers.proxiedsolrapideny.rule=PathRegexp(`^/share/proxy/alfresco(-(noauth|feed|api))?/api/solr/.*$`)"
       - "traefik.http.routers.proxiedsolrapideny.middlewares=fakeauth@docker"
   postgres:
     image: postgres:13.3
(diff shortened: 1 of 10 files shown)


docker-compose/7.2.N-docker-compose.yml
```diff
@@ -119,6 +131,12 @@ services:
         -Dalfresco.port=8080
         -Dalfresco.context=alfresco
         -Dalfresco.protocol=http
+    labels:
+      - "traefik.enable=true"
+      - "traefik.http.routers.share.rule=PathPrefix(`/share`)"
+      - "traefik.http.services.share.loadbalancer.server.port=8080"
+      - "traefik.http.routers.proxiedsolrapideny.rule=PathRegexp(`^/share/proxy/alfresco(-(noauth|feed|api))?/api/solr/`)"
+      - "traefik.http.routers.proxiedsolrapideny.middlewares=fakeauth@docker"
```
pmacius, line 139: Middleware defined in different services can be applied anywhere?

alxgomz: so it seems so though that's strange. Before I disabled the `exposedByDefault` feature of the provisioner I had all middlewares defined at the proxy level and I was just referencing them in the services which needed them (and proxy service was quite at the end of the compose file). After disabling it, traefik started to behave as if the declarations had to be set in the of the file. So to answer your question: Yes you can declare a middleware in some service and reuse later (but not before) with `exposedByDefault` set to `false`. At least that's what I witnessed 🤷

docker-compose/7.1.N-docker-compose.yml
```diff
@@ -59,6 +59,16 @@ services:
...
+    labels:
+      - "traefik.enable=true"
+      - "traefik.http.routers.alfresco.rule=PathPrefix(`/`)"
+      - "traefik.http.services.alfresco.loadbalancer.server.port=8080"
+      - "traefik.http.routers.solrapideny.rule=PathRegexp(`^/alfresco/(wc)?s(ervice)?/api/solr/`)"
+      - "traefik.http.middlewares.fakeauth.basicauth.users=test:"
+      - "traefik.http.routers.solrapideny.middlewares=fakeauth@docker"
+      - "traefik.http.routers.alfrescomicrometer.rule=PathRegexp(`^/alfresco/(wc)?s(ervice)?/prometheus`)"
+      - "traefik.http.middlewares.prometheusipfilter.ipallowlist.sourcerange=127.0.0.0/8"
+      - "traefik.http.routers.alfrescomicrometer.middlewares=prometheusipfilter@docker"
```
gionn, line 71: chatgpt suggests pointing the middleware to a non-existing upstream, plus setting the status explicitly to 403:

```
    labels:
      - "traefik.http.routers.solr-block-router.rule=PathRegexp(`/.*?/service/api/solr/.*`) || PathRegexp(`/.*?/s/api/solr/.*`) || PathRegexp(`/.*?/wcservice/api/solr/.*`) || PathRegexp(`/.*?/wcs/api/solr/.*`) || PathRegexp(`/.*?/proxy/.*/api/solr/.*`) || PathRegexp(`/.*?/-default-/proxy/.*/api/.*`)"
      - "traefik.http.routers.solr-block-router.entrypoints=web"
      - "traefik.http.routers.solr-block-router.middlewares=solr-block-middleware"
      - "traefik.http.middlewares.solr-block-middleware.errorpage.status=403"
      - "traefik.http.middlewares.solr-block-middleware.errorpage.service=noop-service"
      - "traefik.http.middlewares.solr-block-middleware.errorpage.query=/"
      - "traefik.http.services.noop-service.loadbalancer.server.url=http://0.0.0.0"
```

which may be more explicit?

docker-compose/7.1.N-docker-compose.yml
```diff
@@ -59,6 +59,16 @@ services:
...
         -XX:MaxRAM=1900m
+    labels:
+      - "traefik.enable=true"
+      - "traefik.http.routers.alfresco.rule=PathPrefix(`/`)"
+      - "traefik.http.services.alfresco.loadbalancer.server.port=8080"
+      - "traefik.http.routers.solrapideny.rule=PathRegexp(`^/alfresco/(wc)?s(ervice)?/api/solr/`)"
+      - "traefik.http.middlewares.fakeauth.basicauth.users=test:"
+      - "traefik.http.routers.solrapideny.middlewares=fakeauth@docker"
+      - "traefik.http.routers.alfrescomicrometer.rule=PathRegexp(`^/alfresco/(wc)?s(ervice)?/prometheus`)"
+      - "traefik.http.middlewares.prometheusipfilter.ipallowlist.sourcerange=127.0.0.0/8"
```
gionn, line 70: it is sensible that prometheus source ip filter is set to localhost in a container environment?

pmacius (approved):

docker-compose/7.1.N-docker-compose.yml
alxgomz: admitedly not but I can't come up with a sensible default as it's rather random. I see it as a placeholder for someone to easily update it so that the real prometheus IP can be set

docker-compose/7.1.N-docker-compose.yml
alxgomz: both are kinda hacky. I think the one based on `authBasic` middleware is easier to read.
Also I tried and it seems to conflict with the existing router rule as I get errors like:
```
Router alfresco cannot be linked automatically with multiple Services: ["alfresco" "noop-service"] providerName=docker routerName=alfresco
```

pzhyland (approved):

PR #1206, opened 2024-09-30
OPSEXP-2871 Bump tags for Libra release

gionn:
OPSEXP-2871

### DIFF
--- a/docker-compose/pre-release-docker-compose.yml
+++ b/docker-compose/pre-release-docker-compose.yml
@@ -18,7 +18,7 @@
 version: "2"
 services:
   alfresco:
-    image: quay.io/alfresco/alfresco-content-repository:23.4.0-A2
+    image: quay.io/alfresco/alfresco-content-repository:23.4.0-M1
     mem_limit: 1900m
     environment:
       JAVA_TOOL_OPTIONS: >-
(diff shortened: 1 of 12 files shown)


helm/alfresco-content-services/values.yaml
```diff
@@ -555,7 +555,7 @@ dtas:
     assertions:
       acs:
         edition: Enterprise
-        version: 23.3.0
+        version: 23.3.1
```
gionn, line 558: caused by an already seen build issue when doing 23.x patch releases

docker-compose/pre-release-docker-compose.yml
```diff
@@ -197,7 +197,7 @@ services:
       APP_CONFIG_PLUGIN_MICROSOFT_ONLINE: "false"
       APP_BASE_SHARE_URL: "http://localhost:8080/workspace/#/preview/s"
   control-center:
-    image: quay.io/alfresco/alfresco-control-center:9.0.0-10627955718
+    image: quay.io/alfresco/alfresco-control-center:9.1.0-11106981658
```
pmacius, line 200: On quay I next version I see is 10.0.0-11108013258 - we can probably update the matrix for pre release

gionn: matrix is already ok, it's just that I run the workflow before that specific repo started to build new alphas with the bumped version (they are doing it manually after the release happens)

pmacius (approved):

PR #1209, opened 2024-10-02
Fixup HPA integration tests

gionn:


test/autoscaling-hpa-test-values.yaml
```diff
@@ -5,4 +5,4 @@ alfresco-repository:
     maxReplicas: 2
     behavior:
       scaleUp:
-        stabilizationWindowSeconds: 10
+        stabilizationWindowSeconds: 0
```
pmacius, line 8: This is to make sure we dont delay the scaling process?

.github/workflows/helm-enterprise.yml
```diff
@@ -117,8 +117,7 @@ jobs:
         run: |
           if [ "${{ matrix.values }}" = "values.yaml" -o "${{ matrix.values }}" = "pre-release_values.yaml" ]; then
             echo "Enabling clustered tests via auto-scaling: 2 replicas max, 10s stabilization window"
-            echo "helm_install_params=--values test/autoscaling-hpa-test-values.yaml \
-              --set dtas.additionalArgs[0]='-k not test_renditions'" >> "$GITHUB_OUTPUT"
```
pmacius, line 121: We removing this to make sure its overloaded?

pmacius (approved):
Making sure I understand the changes

test/autoscaling-hpa-test-values.yaml
gionn: yes, seems more sensibile for testing even if...

.github/workflows/helm-enterprise.yml
gionn: I think it was a mistake to have this on master, maybe was a leftover from tests

the renditions skip was really necessary for acs 7.0 - and we dropped support for it in the meanwhile

pmacius (approved):

PR #1212, opened 2024-10-02
ACS-8856 Add Audit Storage to the docker compose setup

damianujma:


docker-compose/solr6-override-docker-compose.yml
```diff
@@ -56,6 +56,3 @@ services:
   search-reindexing:
     image: busybox:musl
     command: "true"
-  elasticsearch:
-    image: busybox:musl
```
pmacius, line 60: Why we drop this?

docker-compose/docker-compose.yml
```diff
@@ -207,6 +216,18 @@ services:
       - "5672:5672" # AMQP
       - "61616:61616" # OpenWire
       - "61613:61613" # STOMP
+  audit-storage:
+    image: quay.io/alfresco/alfresco-audit-storage:0.0.1-A7
+    mem_limit: 1g
```
pmacius, line 221: Can it be lower?

docker-compose/solr6-override-docker-compose.yml
damianujma: Audit Storages requires Elasticsearch to store audit logs. In that case, we don't need to disable Elasticsearch in the solr6 docker-compose file.

docker-compose/docker-compose.yml
damianujma: Reduced to `512m`.

docker-compose/docker-compose.yml
```diff
@@ -207,6 +216,18 @@ services:
...
+    mem_limit: 512m
+    environment:
+      SPRING_ACTIVEMQ_BROKERURL: failover:(nio://activemq:61616)?timeout=3000
+      AUDIT_ENTRYSTORAGE_OPENSEARCH_CONNECTOR_URI: http://elasticsearch:9200
+      AUDIT_ENTRYSTORAGE_OPENSEARCH_CONNECTOR_USERNAME: admin
+      AUDIT_ENTRYSTORAGE_OPENSEARCH_CONNECTOR_PASSWORD: admin
+      AUDIT_EVENTINGESTION_URI: activemq:topic:alfresco.repo.event2
+    depends_on:
+      - activemq
+      - elasticsearch
```
gionn, line 230: @alxgomz we should probably move not-yet-released components to pre-release yml instead?

not sure if this new component should be treated as an optional one - so we could have a separate file like we did with solr6

gionn (changes requested):

docker-compose/docker-compose.yml
pzhyland: This component will be released "soon" The main goal of this PR is to get your opinion, adjust the config and be ready for the GA.

We have most of our components in the main docker-compose file so I think the audit storage should be there as well.

docker-compose/docker-compose.yml
```diff
@@ -146,7 +146,7 @@ services:
     ports:
       - "5432:5432"
   elasticsearch:
-    image: elasticsearch:7.10.1
+    image: elasticsearch:7.17.3
```
gionn, line 149: bump it also in pre-release

docker-compose/docker-compose.yml
```diff
@@ -162,6 +162,15 @@ services:
     ports:
       - 9200:9200
       - 9300:9300
+  kibana:
+    image: kibana:7.17.3
+    mem_limit: 1g
```
gionn, line 167: for sure kibana doesn't need `1g`, `100m`?

gionn: looks like things changed a lot since last time I played with it 😬 

helm charts defaults to 2g:
https://github.com/elastic/helm-charts/blob/main/kibana/values.yaml#L55

so if it plays well with 1g for you let's stick to that

docker-compose/docker-compose.yml
alxgomz, line 149: Is that really required? I'd rather we stick to an image which embeds the opensource version of ES. I have not checked what the real implications of using the new ES license are. But I'm not planning to do so and I'd rather stay on the safe side.

docker-compose/docker-compose.yml
alxgomz: pre-release imho, and we'll need to update the updatecli matrix as well.
Now, what worries me a bit here is to have that enabled by default. running this docker compose file will soon require more than 8GB and that's the base for the download trial. So maybe we should have an override file for this (and a related test job)?
Is there a frontend to that new feature already? Something trial users would really see a benefit for when testing Alfresco?

docker-compose/docker-compose.yml
damianujma: We need to have the same version for kibana and elasticsearch. There is no ARM64 image for `kibana:7.10.1`.

gionn: the elastic license should not really be an issue for us shipping reference docker compose configs, see https://www.elastic.co/licensing/elastic-license/faq

docker-compose/docker-compose.yml
gionn: I guess the frontend is kibana itself, customized dashboard are still under development, eventually we need to ship those as well

alxgomz (approved):

pzhyland (approved):

PR #1218, opened 2024-10-04
OPSEXP-2881: trigger docs publication on content change on master and releases

alxgomz:
Ref: OPSEXP-2881

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -6,6 +6,9 @@ permissions:
   id-token: write
 
 on:
+  release:
+    types:
+      - released
   push:
     branches:
       - master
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -6,6 +6,9 @@ permissions:
   id-token: write
 
 on:
+  release:
+    types:
+      - released
   push:
     branches:
       - master
```
gionn, line 14: I would fix the paths anyway, we did the same already for alfresco-helm-charts https://github.com/Alfresco/alfresco-helm-charts/blob/main/.github/workflows/docs.yml#L17

```
- helm/**/README.md
- helm/**/docs/**.md
```

also consider that this workflow will run twice in a short-time if the last PR merged before the release contains changes in `docs/**` - but yeah should not happen so often

alxgomz: Surely if I add these additional paths the wf will always run twice on release, as the charts READMEs will always be updated. I think it's preferable to not include them. All in all, these files are part of the chart and a change in these are worth a release eventually.

gionn (approved):

pmacius (approved):

PR #1222, opened 2024-10-07
Add community-specific dtas configuration to integration values

gionn:
basic assertions would fail otherwise with community edition

### DIFF
--- a/test/community-integration-test-values.yaml
+++ b/test/community-integration-test-values.yaml
@@ -92,3 +92,8 @@ postgresql:
         memory: "1Gi"
 dtas:
   enabled: true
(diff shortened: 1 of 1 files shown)


test/community-integration-test-values.yaml
```diff
@@ -92,3 +92,8 @@ postgresql:
         memory: "1Gi"
 dtas:
   enabled: true
+  config:
+    assertions:
+      acs:
+        edition: Community
+      adw: null
```
pmacius, line 99: Is it possible to override with null now?

pmacius (approved):

PR #1226, opened 2024-10-15
OPSEXP-2878 Add alfresco-audit-storage chart

pmacius:
Ref: OPSEXP-2878

### DIFF
--- a/helm/alfresco-content-services/Chart.yaml
+++ b/helm/alfresco-content-services/Chart.yaml
@@ -4,7 +4,7 @@
 # https://www.alfresco.com/platform/content-services-ecm/trial/download
 apiVersion: v2
 name: alfresco-content-services
-version: 8.5.2
+version: 8.6.0-alpha.0
 appVersion: 23.3.2
 description: A Helm chart for deploying Alfresco Content Services
 keywords:
--- a/helm/alfresco-content-services/Chart.yaml
+++ b/helm/alfresco-content-services/Chart.yaml
@@ -85,6 +85,11 @@ dependencies:
     repository: https://helm.elastic.co
     version: 7.17.3
     condition: elasticsearch.enabled
+  - name: elasticsearch
+    alias: elasticsearchAas
+    repository: https://helm.elastic.co
+    version: 7.17.3
+    condition: elasticsearchAas.enabled
   - name: kibana
     repository: https://helm.elastic.co
     version: 7.17.3
--- a/helm/alfresco-content-services/Chart.yaml
+++ b/helm/alfresco-content-services/Chart.yaml
@@ -91,10 +91,10 @@ dependencies:
     version: 7.17.3
     condition: elasticsearchAudit.enabled
   - name: kibana
-    alias: kibanaAudit
+    alias: kibana-audit
     repository: https://helm.elastic.co
     version: 7.17.3
-    condition: kibanaAudit.enabled
+    condition: kibana-audit.enabled
   - name: alfresco-audit-storage
     version: 0.1.0-alpha.0
     repository: https://alfresco.github.io/alfresco-helm-charts/
--- a/helm/alfresco-content-services/Chart.yaml
+++ b/helm/alfresco-content-services/Chart.yaml
@@ -86,10 +86,10 @@ dependencies:
     version: 7.17.3
     condition: elasticsearch.enabled
   - name: elasticsearch
-    alias: elasticsearchAudit
+    alias: elasticsearch-audit
     repository: https://helm.elastic.co
     version: 7.17.3
-    condition: elasticsearchAudit.enabled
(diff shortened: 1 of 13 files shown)


helm/alfresco-content-services/values.yaml
```diff
@@ -538,6 +551,70 @@ elasticsearch:
...
+  clusterHealthCheckParams: "wait_for_status=yellow&timeout=1s"
+  clusterName: elasticsearch-aas
+  ingress:
+    enabled: true
+    annotations: 
+      nginx.ingress.kubernetes.io/rewrite-target: /$2
+    hosts:
+      - paths:
+          - path: /elasticsearch(/|$)(.*)
+            pathType: Prefix
```
gionn, line 568: we can't expose elasticsearch via ingress directly as it doesn't have any authentication, is that mandatory for kibana?

helm/alfresco-content-services/Chart.yaml
```diff
@@ -85,4 +85,17 @@ dependencies:
     repository: https://helm.elastic.co
     version: 7.17.3
     condition: elasticsearch.enabled
+  - name: elasticsearch
+    alias: elasticsearchAudit
+    repository: https://helm.elastic.co
+    version: 7.17.3
+    condition: elasticsearchAudit.enabled
```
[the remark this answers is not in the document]
pmacius: We discussed with @alxgomz that it should have its own elasticsearch this way we don't overcomplicate the deployment with solr6 but I can't recall what was the main reason.

helm/alfresco-content-services/values.yaml
pzhyland: It's correct. The audit storage doesn't have any UI/API and we need to access the ES from tests to verify everything works correctly. @cezary-witkowski created a test env where ES is protected with basic auth and the password is taken from the pipeline secrets.

gionn: for sure we can't have here a default configuration which is insecure, it's fine that we enabled it on internal environments behind basic auth

alxgomz (changes requested):
Can you also update the other value files?

helm/alfresco-content-services/values.yaml
alxgomz: otherwise if it's needed just for tests let's have ES not exposed , and components coudl use the service directly, and we add the ingress only in TF pipeline?

helm/alfresco-content-services/Chart.yaml
```diff
@@ -85,4 +85,18 @@ dependencies:
     repository: https://helm.elastic.co
     version: 7.17.3
     condition: elasticsearch.enabled
+  - name: elasticsearch
+    alias: elasticsearchAudit
```
alxgomz, line 89: I'll be extra painful here but this alias is the only one using camelCase. Please switch it to `elasticsearch-audit`

helm/alfresco-content-services/values.yaml
```diff
@@ -538,6 +551,91 @@ elasticsearch:
...
+      valueFrom:
+        secretKeyRef: 
+          name: *aas_elasticsearch_secretName
+          key: AUDIT_ELASTICSEARCH_USERNAME
+    - name: ELASTIC_PASSWORD
+      valueFrom:
+        secretKeyRef: 
+          name: *aas_elasticsearch_secretName
+          key: AUDIT_ELASTICSEARCH_PASSWORD
+  ingress:
```
alxgomz, line 572: The ingress is disabled by default, but that's a good thing to have it explicitly disabled here. Though further config is not needed here. We should maybe just link to the upstream chart doc for ingress config details:

alxgomz (changes requested):

damianujma (approved):

alxgomz (approved):

kmagdziarz (approved):

helm/alfresco-content-services/values.yaml
```diff
@@ -538,6 +551,87 @@ elasticsearch:
...
+      name: *aas_elasticsearch_secretName
+      keys:
+        username: AUDIT_ELASTICSEARCH_USERNAME
+        password: AUDIT_ELASTICSEARCH_PASSWORD
+kibana-audit:
+  enabled: true
+  healthCheckPath: "/kibana/app/kibana"
+  # -- Makes sure there is no default elasticsearch hosts defined
+  elasticsearchHosts: ""
+  # -- All of the values has to be set there to escape the issue with overriding the values
```
gionn, line 603:
```suggestion
  # All of the values has to be set there to escape the issue with overriding the values
```

see if the readme improve - in the end this is not really a comment for the chart end user

gionn (approved):

gionn (approved):

PR #1241, opened 2024-10-30
MVP 7.4 and 23.3 with Elastic Search

todorinskiz:


todorinskiz:
pushed it accidentally to the source repo instead of the fork, apologies

PR #1243, opened 2024-11-04
OPSEXP-2694: automatic updates of acs sso example chart

alxgomz:
Ref: OPSEXP-2694

I didn't to have different updatecli pipelines in the first place, though merging the compatibility matrix values with the local matrix of targets using yq would require merging by subkey instead of key name which seems overly complicated.
Requires https://github.com/Alfresco/alfresco-updatecli/pull/44

Tested in:
* https://github.com/Alfresco/acs-deployment/actions/runs/11684259267/job/32535136474
* https://github.com/Alfresco/acs-deployment/actions/runs/11682116291


helm/alfresco-content-services/Chart.yaml
```diff
@@ -93,7 +93,7 @@ dependencies:
   - name: kibana
     alias: kibana-audit
     repository: https://helm.elastic.co
-    version: 7.17.3
+    version: 8.5.1
```
pmacius, line 96: Not sure about bumping this chart. Maybe we should stay on the same version as elasticsearch but considering it doesnt break anything in workflow it should be fine. Im more worried about pipeline envs

updatecli.d/helm-autodiscovery.yaml
```diff
@@ -17,5 +17,6 @@ autodiscovery:
           dependencies:
             postgresql: ">0.0.1"
             elasticsearch: ">0.0.1"
```
pmacius, line 19: maybe we should add kibana here

.github/workflows/bumpVersions.yml
```diff
@@ -92,16 +92,14 @@ jobs:
...
-          tee ./merged.yaml
-
-      - name: updatecli apply
-        # path to values file must be relative https://github.com/updatecli/updatecli/issues/1253
-        run: updatecli apply -c alfresco-updatecli/deployments/uber-manifest.tpl -v ./merged.yaml
+          for i in {docker-compose,helm/*}/updatecli-matrix-targets.yaml; do TARGET=$(basename $(dirname $i))
+             echo -e "\n###### Building UpdateCli manifest for ${TARGET}\n"
+             yq '. *? load("alfresco-updatecli/deployments/values/supported-matrix.yaml") | explode(.)' ${i} | tee ./${TARGET}_merged.yaml
+             updatecli apply -c alfresco-updatecli/deployments/uber-manifest.tpl -v ./${TARGET}_merged.yaml
+          done
```
gionn, line 102:
```suggestion
          for i in {docker-compose,helm/*}/updatecli-matrix-targets.yaml; do 
            TARGET=$(basename $(dirname $i))
            echo -e "\n###### Building Updatecli values for ${TARGET}\n"
            yq '. *? load("alfresco-updatecli/deployments/values/supported-matrix.yaml") | explode(.)' ${i} | tee ./${TARGET}_merged.yaml
            updatecli apply -c alfresco-updatecli/deployments/uber-manifest.tpl -v ./${TARGET}_merged.yaml
          done
```

.github/workflows/bumpVersions.yml
```diff
@@ -92,16 +92,14 @@ jobs:
           ref: ${{ inputs.alfresco-updatecli-ref || env.DEFAULT_BRANCH_NAME }}
           path: alfresco-updatecli
 
-      - name: Preprocess values file appending existing keys only
+      - name: Build manifest and run UpdateCli pipelines
```
gionn, line 95:
```suggestion
      - name: Build values and run Updatecli pipelines
```

docker-compose/7.1.N-compose.yaml
```diff
@@ -194,15 +194,15 @@ services:
     mem_limit: 2g
     environment:
       # Solr needs to know how to register itself with Alfresco
-      - SOLR_ALFRESCO_HOST=alfresco
-      - SOLR_ALFRESCO_PORT=8080
+      SOLR_ALFRESCO_HOST: alfresco
+      SOLR_ALFRESCO_PORT: 8080
```
gionn, line 198: why? the most current notation is with the equal sign

helm/acs-sso-example/.helmignore
```diff
@@ -21,3 +21,5 @@
 .idea/
 *.tmproj
 .vscode/
+# Do not include UpdateCli target matrix
```
gionn, line 24:
```suggestion
# Do not include Updatecli target matrix
```

helm/acs-sso-example/Chart.yaml
```diff
@@ -19,7 +19,6 @@ description: |
 
   :warning: All components have persistence disabled so all data is lost after a
   deployment is destroyed or rolled back!
-
 type: application
 version: 1.0.0
```
gionn, line 23: bump it - alpha?

helm/acs-sso-example/Chart.yaml
```diff
@@ -19,7 +19,6 @@ description: |
...
 appVersion: 23.2.2
```
gionn, line 24: is this in sync?

helm/acs-sso-example/values.yaml
```diff
@@ -73,8 +71,8 @@ keycloakx:
             webOrigins: >-
               {{ index (include "alfresco-common.known.urls" $ | mustFromJson) "known_urls" }}
         users:
-          -  # -- default Alfresco admin user
-            username: admin
+          # -- default Alfresco admin user
+          - username: admin
```
gionn, line 75: syntax may look odd but I think it was intended to have the doc entry only for the username field, now it is attached to the entire object

helm/alfresco-content-services/.helmignore
```diff
@@ -23,3 +23,5 @@ tests
 .idea/
 *.tmproj
 .vscode/
+# Do not include UpdateCli target matrix
```
gionn, line 26:
```suggestion
# Do not include Updatecli target matrix
```

helm/alfresco-content-services/Chart.lock
```diff
@@ -52,9 +52,9 @@ dependencies:
   version: 7.17.3
 - name: kibana
   repository: https://helm.elastic.co
-  version: 7.17.3
+  version: 8.5.1
```
gionn, line 55: check

docker-compose/7.1.N-compose.yaml
gionn: of course I just looked at postgres 😀 now I remember that kics was warning about that  but only for dockerfiles.

at this point if postgres is really the last one you could align that as well

gionn (approved):

PR #1264, opened 2024-12-03
OPSEXP-2961 Disable services for solr6 overrides using reset yaml tag and bump pre-commit repos versions

pmacius:
Ref: OPSEXP-2961

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,8 +42,7 @@ repos:
     rev: v2.1.1
     hooks:
       - id: docker-compose-check
-        files : docker-compose/.*yaml
-        exclude: docker-compose/solr6-overrides.yaml
+        files : docker-compose/.*compose.yaml
 
   - repo: local
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,7 +42,7 @@ repos:
     rev: v2.1.1
     hooks:
       - id: docker-compose-check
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -42,7 +42,20 @@ repos:
     rev: v2.1.1
```
gionn, line 42: try also to bump to latest version before

.pre-commit-config.yaml
```diff
@@ -42,7 +42,20 @@ repos:
...
-        exclude: solr6-overrides.yaml
+        files : docker-compose/.*compose.yaml
+
+  - repo: local
+    hooks:
+      - id: check-docker-compose-solr6-overrides
+        name: Check Docker Compose Solr6 Overrides
+        entry: |
+          bash -c "
+          docker compose -f docker-compose/compose.yaml -f docker-compose/solr6-overrides.yaml config |
```
gionn, line 53: isn't the same thing done by the upstream hook? https://github.com/IamTheFij/docker-pre-commit/blob/master/compose-check.sh#L28

.pre-commit-config.yaml
```diff
@@ -42,7 +42,20 @@ repos:
...
+          yamllint --config-data '{extends: default, rules: {line-length: disable, document-start: {present: false}}}' -
```
gionn, line 54: what are you overriding here which you can't override in the main yamllint hook?

gionn (approved):

PR #1267, opened 2024-12-09
OPSEXP-2950 Bump elasticsearch to 8.16.1 for pre-release

pmacius:
Ref: OPSEXP-2950

### DIFF
--- a/docker-compose/pre-release-compose.yaml
+++ b/docker-compose/pre-release-compose.yaml
@@ -181,7 +181,7 @@ services:
       retries: 3
       start_period: 5s
   elasticsearch:
-    image: elasticsearch:7.17.3
+    image: elasticsearch:8.16.1
     environment:
       - xpack.security.enabled=false
       - discovery.type=single-node
@@ -204,7 +204,7 @@ services:
       retries: 5
       start_period: 30s
   kibana:
-    image: kibana:7.17.3
+    image: kibana:8.16.1
     mem_limit: 1g
     environment:
       - ELASTICSEARCH_HOSTS=http://elasticsearch:9200
--- a/helm/alfresco-content-services/pre-release_values.yaml
+++ b/helm/alfresco-content-services/pre-release_values.yaml
@@ -78,9 +78,13 @@ alfresco-connector-ms365:
 alfresco-connector-msteams:
   image:
     tag: 2.0.4-A4
+elasticsearch-audit:
+  imageTag: 8.16.1
 alfresco-audit-storage:
   image:
     tag: 1.0.0
+elasticsearch: 
+  imageTag: 8.16.1
 dtas:
   config:
     assertions:


helm/alfresco-content-services/pre-release_values.yaml
```diff
@@ -78,6 +78,12 @@ alfresco-connector-ms365:
 alfresco-connector-msteams:
   image:
     tag: 2.0.4-A4
+elastisearch:
+  image:
+    tag: 8.17.0
```
alxgomz, line 83: At that point of repo is not ready let's stick to 7

helm/alfresco-content-services/templates/config-infrastructure.yaml
```diff
@@ -28,11 +28,11 @@ data:
     {{- else if eq "solr6" $search_flavor }}
       {{- $search_url = printf "http://%s/solr" (include "alfresco-search-service.deployment.name" .) }}
     {{- else if eq "elasticsearch" $search_flavor }}
+      {{- $esHost := printf "%s-%s" (.Release.Name | default "acs") (.Values.global.elasticsearch.service.name | default "elasticsearch") }}
+      {{- $esPort := .Values.global.elasticsearch.service.ports.restApi | default 9200 }}
```
alxgomz, line 32: I'm not sure why this has been moved outside of the `if .enabled`.

helm/alfresco-content-services/templates/config-infrastructure.yaml
```diff
@@ -73,18 +73,17 @@ data:
     {{- if .Values.global.auditIndex.url }}
       {{- $elasticsearch_audit_url = .Values.global.auditIndex.url }}
     {{- else }}
-      {{- with (index .Values "elasticsearch-audit") }}
+      {{- $auditEsHost := printf "%s-%s" (.Release.Name | default "acs") (.Values.global.elasticsearch.service.name | default "elasticsearch") }}
```
alxgomz, line 76: that default (`acs`) looks like a pipeline specific. I don't think we need to default `Release.Name`

I have the same remark that I think it should remain inside the `if .enabled` statement below

helm/alfresco-content-services/tests/search_test.yaml
```diff
@@ -175,3 +176,67 @@ tests:
...
+  - it: Should set correct credentials with external audit and external elastisearch as the same cluster
+    values: *testvalues
+    set:
+      global:
+        search:
+          username: elastic
+          password: changeme
+        auditIndex:
+          username: elastic
+          password: changeme
```
alxgomz, line 225: maybe use different creds in the test so it doesn't accidentally pass in case we mistakenly swap them somehow

helm/alfresco-content-services/values.yaml
```diff
@@ -63,18 +63,22 @@ global:
...
     secretName: &aas_elasticsearch_secretName alfresco-aas-elasticsearch-secret
     # -- Name of an existing secret that contains AUDIT_ELASTICSEARCH_USERNAME and AUDIT_ELASTICSEARCH_PASSWORD keys.
     existingSecretName: null
-    # -- Base url for kibana environment variable `SERVER_PUBLICBASEURL` 
-    publicBaseUrl: null
+  elasticsearch:
+    service:
+      name: elasticsearch
+      ports:
+        restAPI: 9200
```
alxgomz, line 80: This being part of the global namespace it should not be needed and be available IF people DO enable the chart.
Can you test it?

alxgomz (changes requested):

helm/alfresco-content-services/templates/config-infrastructure.yaml
pmacius: I'll move it back and use `$`

helm/alfresco-content-services/values.yaml
```diff
@@ -63,18 +63,17 @@ global:
     # is solr6 or SEARCH_USERNAME and SEARCH_PASSWORD keys.
     existingSecretName: null
   auditIndex:
-    # -- Elasticsearch URL
+    # -- set this URL if you have an external search service for audit indexing
```
alxgomz, line 66:
```suggestion
    # -- set this URL if you have an external elasticsearch for audit indexing
```

helm/alfresco-content-services/values.yaml
```diff
@@ -662,7 +628,7 @@ dtas:
         base_path: "/workspace"
       aas:
         audit_host: http://acs-alfresco-audit-storage:8081
-        elasticsearch_host: http://elasticsearch-aas-master:9200
+        elasticsearch_host: http://acs-elasticsearch:9200
```
alxgomz, line 631: this will fail in case the release name is not "acs". We should probably turn dtas.config into a string and use a tpl() function to render the dtas config json

helm/alfresco-content-services/values.yaml
```diff
@@ -549,29 +548,33 @@ alfresco-ai-transformer:
...
+    replicaCount: 0
+  ingest:
+    replicaCount: 0
+  kibana:
+    image:
+      tag: 7.17.26
+    configuration:
+      server:
+        basePath: "/kibana"
+        publicBaseUrl: "http://localhost/kibana"
```
alxgomz, line 568: That looks a bit odd to me. Did you check wether kibana was properly accessible in integration pipeline?

docker-compose/pre-release-compose.yaml
alxgomz, line 1: I don't know if it works because we're not testing with dtas but it looks to me like it should not be different from the helm pre-release values

alxgomz (changes requested):

helm/alfresco-content-services/values.yaml
gionn: in global.known_urls we assume default deployment target localhost, so at least is consistent

just add more inline docs to make it easy to understand that it must be set for other types of deployment

helm/alfresco-content-services/values.yaml
gionn: we are not really doing this for other service-discovery related values, I would not start doing it for dtas which is quite optional now.

maybe configure by default the service to not include the release name?

test/enterprise-integration-test-values.yaml
gionn, line 1: check also the examples values file under docs if it need some adjustement - maybe we should mention it here in a comment to remember about it xD

helm/alfresco-content-services/values.yaml
```diff
@@ -661,8 +633,8 @@ dtas:
       adw:
         base_path: "/workspace"
       aas:
-        audit_host: http://acs-alfresco-audit-storage:8081
-        elasticsearch_host: http://elasticsearch-aas-master:9200
+        audit_host: '{{ include "alfresco-content-services.aasService" $ }}'
```
gionn, line 636: alfresco-content-services.audit.serviceName?

helm/alfresco-content-services/values.yaml
```diff
@@ -595,46 +604,9 @@ alfresco-audit-storage:
...
-  ingress:
-    enabled: true
-    hosts:
-      - paths:
-          - path: /kibana
-            pathType: Prefix
 dtas:
   # -- Enables the deployment test suite which can run via `helm test` (currently available for Enterprise only)
-  enabled: false
+  enabled: true
```
gionn, line 609: explain

helm/alfresco-content-services/values.yaml
```diff
@@ -549,32 +548,42 @@ alfresco-ai-transformer:
...
+  master:
+    masterOnly: false
+    replicaCount: 1
+  data:
+    replicaCount: 0
+  coordinating:
+    replicaCount: 0
+  ingest:
+    replicaCount: 0
+  kibana:
```
gionn, line 562: this kibana supports any authentication?

helm/alfresco-content-services/README.md
```diff
@@ -237,32 +236,33 @@ Please refer to the [documentation](https://github.com/Alfresco/acs-deployment/b
...
 | global.auditIndex.existingSecretName | string | `nil` | Name of an existing secret that contains AUDIT_ELASTICSEARCH_USERNAME and AUDIT_ELASTICSEARCH_PASSWORD keys. |
-| global.auditIndex.password | string | `nil` | Elasticsearch password |
-| global.auditIndex.publicBaseUrl | string | `nil` | Base url for kibana environment variable `SERVER_PUBLICBASEURL`  |
+| global.auditIndex.password | string | `nil` | set password for authentication against the external elasticsearch service for audit indexing |
 | global.auditIndex.secretName | string | `"alfresco-aas-elasticsearch-secret"` | Name of the secret managed by this chart |
-| global.auditIndex.url | string | `nil` | Elasticsearch URL |
-| global.auditIndex.username | string | `nil` | Elasticsearch username |
+| global.auditIndex.url | string | `nil` | set this URL if you have an external elasticsearch for audit indexing |
+| global.auditIndex.username | string | `nil` | set usernname for authentication against the external elasticsearch service for audit indexing |
+| global.kibanaEnabled | bool | `true` |  |
```
gionn, line 265: add doc

helm/alfresco-content-services/values.yaml
pmacius: Sorry leftover

alxgomz (commented):
OK for me with Kibana disabled by default and a small doc on how to enable and secure it leveraging mechanisms provided by the bitnami chart

helm/alfresco-content-services/values.yaml
pmacius: added docs

docs/helm/examples/with-elasticsearch-auth.md
alxgomz, line 1: We should make it clear this is in case one uses the embedded elasticsearch which is not recommended for prod

alxgomz (approved):

docs/helm/examples/with-elasticsearch-auth.md
```diff
@@ -0,0 +1,77 @@
...
+ELASTIC_USERNAME="elastic" ELASTIC_PASSWORD="alfresco" KIBANA_PASSWORD="alfrescokibana" && \
+kubectl create secret generic elastic-search-secret \
+    --namespace=default \
+    --from-literal=elastic-user=$ELASTIC_USERNAME \
+    --from-literal=elasticsearch-password=$ELASTIC_PASSWORD \
+    --from-literal=kibana-password=$KIBANA_PASSWORD \
+    --from-literal=AUDIT_ELASTICSEARCH_USERNAME=$ELASTIC_USERNAME \
+    --from-literal=AUDIT_ELASTICSEARCH_PASSWORD=$ELASTIC_PASSWORD \
+    --from-literal=SEARCH_USERNAME=$ELASTIC_USERNAME \
+    --from-literal=SEARCH_PASSWORD=$ELASTIC_PASSWORD
```
gionn, line 43: see `--from-env-file` which should prevent user leaking credentials in its history

gionn (approved):

PR #1278, opened 2025-01-20
OPSEXP-3005 Document using specific version of ingress-nginx to avoid breakage

pmacius:
Ref: OPSEXP-3005 
https://github.com/kubernetes/ingress-nginx/issues/12618

They changed the behavior of `{"allow-snippet-annotations":"true"}}'` - it seems like it doesn't work anymore
Currently even with this set to true we will get: 
```
        * admission webhook "validate.nginx.ingress.kubernetes.io" denied the request: annotation group ServerSnippet contains risky annotation based on ingress configuration
        * admission webhook "validate.nginx.ingress.kubernetes.io" denied the request: annotation group ServerSnippet contains risky annotation based on ingress configuration
```
It is required to change it to ` {"annotations-risk-level":"Critical"}}' `

### DIFF
--- a/docs/helm/ingress-nginx.md
+++ b/docs/helm/ingress-nginx.md
@@ -31,7 +31,7 @@ kubectl -n ingress-nginx patch cm ingress-nginx-controller \
 ```
 
 :warning: For latests versions of nginx it is required to use
-`"annotations-risk-level":"Critical"`, see: this
+`"annotations-risk-level":"Critical"`, see this
 [issue](https://github.com/kubernetes/ingress-nginx/issues/12618)
 
 Wait for the ingress-nginx controller to be up again after the configuration change:
--- a/docs/helm/ingress-nginx.md
+++ b/docs/helm/ingress-nginx.md
@@ -18,8 +18,11 @@ Install the ingress-nginx controller namespace:
 ```bash
 helm upgrade --install ingress-nginx ingress-nginx \
 --repo https://kubernetes.github.io/ingress-nginx \
---namespace ingress-nginx --create-namespace
+--namespace ingress-nginx --create-namespace \
+--version 4.7.2
 ```
+
+### Ingress configmap patch
 
 Enable snippet annotations which is disabled by default for security reasons, but
 we still requires it for `alfresco-search-services` while still filtering only
@@ -27,10 +30,10 @@ the ones we strictly need.
 
 ```bash
 kubectl -n ingress-nginx patch cm ingress-nginx-controller \
--p '{"data": {"annotations-risk-level":"Critical", "allow-snippet-annotations":"true"}}'
+-p '{"data": {"allow-snippet-annotations":"true"}}'
 ```
 
-:warning: Since nginx controller v1.12 or ingress-nginx helm chart v4.12 use
+:warning: Since nginx controller v1.12 or ingress-nginx helm chart v4.12.0 use
 `"annotations-risk-level":"Critical"`. For versions prior to that use
 `"allow-snippet-annotations":"true"`
 
--- a/docs/helm/ingress-nginx.md
+++ b/docs/helm/ingress-nginx.md
@@ -24,9 +24,12 @@ helm upgrade --install ingress-nginx ingress-nginx \
 
 ### Ingress configmap patch
 
-Enable snippet annotations which is disabled by default for security reasons, but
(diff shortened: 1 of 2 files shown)


docs/helm/kind-deployment.md
```diff
@@ -64,7 +64,7 @@ when using our search services chart:
 
 ```sh
 kubectl -n ingress-nginx patch cm ingress-nginx-controller \
-  -p '{"data": {"allow-snippet-annotations":"true"}}'
+  -p '{"data": {"annotations-risk-level":"Critical"}}'
```
gionn, line 67: keep both so it works for every version

docs/helm/kind-deployment.md
```diff
@@ -56,17 +56,26 @@ Wait for the Kind cluster to be created. This may take a few minutes.
...
 
-Follow the instructions for
-[deploying nginx ingress on KinD](https://kind.sigs.k8s.io/docs/user/ingress/#ingress-nginx).
+Install the ingress-nginx controller namespace:
+
+```bash
+helm upgrade --install ingress-nginx ingress-nginx \
+--repo https://kubernetes.github.io/ingress-nginx \
+--namespace ingress-nginx --create-namespace
+```
```
gionn, line 65: see https://github.com/kubernetes-sigs/kind/issues/1693#issuecomment-1768116494

I don't expect to be working without any change, otherwise there was no need to have a "kind" flavor of ingress-nginx

docs/helm/kind-deployment.md
```diff
@@ -56,17 +56,26 @@ Wait for the Kind cluster to be created. This may take a few minutes.
...
 ```sh
 kubectl -n ingress-nginx patch cm ingress-nginx-controller \
-  -p '{"data": {"allow-snippet-annotations":"true"}}'
+  -p '{"data": {"annotations-risk-level":"Critical", "allow-snippet-annotations":"true"}}'
 ```
 
+:warning: For latest versions of nginx it is required to use
+`"annotations-risk-level":"Critical"`, see this
+[issue](https://github.com/kubernetes/ingress-nginx/issues/12618)
+
```
gionn, line 78: I think it's less confusing for the user if we say: since version X use this, before use that - no link to the issue given it's not an issue and it's going to stay like this for upcoming versions

docs/helm/ingress-nginx.md
gionn, line 1: add an explicit version to helm install command so it's consistent with kind

another idea could be to add a title for the ingress-patch so you can link it from kind and no duplication in the docs.

gionn (changes requested):

docs/helm/ingress-nginx.md
```diff
@@ -45,5 +37,15 @@ Verify the newly created pod under the ingress-nginx namespace:
 kubectl get pods --namespace=ingress-nginx
 ```
 
+### Optional patching
```
gionn, line 40:
```suggestion
### Enabling ingress annotations
```

or something more specific, it's not optional if it's required for the charts to work

docs/helm/ingress-nginx.md
```diff
@@ -45,5 +37,15 @@ Verify the newly created pod under the ingress-nginx namespace:
...
+
+Since nginx controller v1.12 or ingress-nginx helm chart v4.12.0 use
+`"annotations-risk-level":"Critical"`.
```
gionn, line 43:
```suggestion
Since nginx controller v1.12 or ingress-nginx helm chart v4.12.0 it become mandatory to allow annotations on Ingress.
```

docs/helm/kind-deployment.md
```diff
@@ -54,27 +54,16 @@ EOF
...
-when using our search services chart:
-
-```sh
-kubectl -n ingress-nginx patch cm ingress-nginx-controller \
-  -p '{"data": {"allow-snippet-annotations":"true"}}'
+```bash
+kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/controller-v1.8.2/deploy/static/provider/kind/deploy.yaml
 ```
 
-Wait for the ingress-nginx controller to be up again after the configuration change:
```
gionn, line 70: keep it, it still good to have an health check before proceeding

gionn (approved):

gionn (approved):

PR #1283, opened 2025-01-28
OPSEXP-2917 Docs: remove unnecessary Solr shared secret for enterprise

pmacius:
ref: OPSEXP-2917

### DIFF
--- a/docs/helm/desktop-deployment.md
+++ b/docs/helm/desktop-deployment.md
@@ -130,6 +130,24 @@ helm install acs alfresco/alfresco-content-services \
 
 The `helm` command above installs the most current released version of ACS Enterprise.
 
+#### Enterprise with solr6
+
+Use the above helm command with additional arguments that enable solr6 search engine on enterprise.
+
+```bash
+helm upgrade acs helm/alfresco-content-services \
+  --values docs/helm/values/local-dev_values.yaml \
+  --set alfresco-repository.configuration.search.flavor=solr6 \
+  --set global.search.sharedSecret=$(openssl rand -hex 24) \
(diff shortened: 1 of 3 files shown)


gionn (commented):
for community search service is still the default and that's ok - however I don't see anywhere where we tell the user how to install enterprise with search service, we should have a short paragraph somewhere

docs/helm/desktop-deployment.md
```diff
@@ -130,6 +130,24 @@ helm install acs alfresco/alfresco-content-services \
...
+
+```bash
+helm upgrade acs helm/alfresco-content-services \
+  --values docs/helm/values/local-dev_values.yaml \
+  --set alfresco-repository.configuration.search.flavor=solr6 \
+  --set global.search.sharedSecret=$(openssl rand -hex 24) \
+  --set alfresco-search-enterprise.enabled=false \
+  --set alfresco-search.enabled=true \
+  --set elasticsearch.enabled=false \
+  --set alfresco-audit-storage.enabled=false \
```
gionn, line 145: in the end I am not 100% sure, but at least will prevent user confused by getting an elasticsearch cluster anyway

alxgomz (approved):

docs/helm/examples/with-elasticsearch-auth.md
```diff
@@ -61,7 +61,6 @@ Deploy the ACS stack with the appropriate values files.
 
 ```bash
 helm install acs ./helm/alfresco-content-services \
```
gionn, line 63: here we are assuming user using git but most of other examples use the chart registry, they should be coherent.

this would make tricky to fetch the additional required values file, we should also improve that

PR #1301, opened 2025-02-26
OPSEXP-3109 Provide anonymous volumes for alfresco content store and sfs

### LINKED ISSUE #649: Support persistent data when restarting ACS in Docker Compose deployment
Hello together,

for evaluation purposes, I wanted to test your images to deploy Alfresco via Docker.
Since we use Docker-Stack for production, we would also like to use Alfresco with stack.
After a few setup hurdles, it basically works quite well.

However we have a massive problem when restarting the services.
The containers start, but the following appears on the root URL:
<pre>
Cannot find Alfresco Repository on this server. (Does this application have access to alfresco-global.properties? Does this application have cross-context permissions?)
</pre>
In addition, the share and all other components are no longer usable.
In posts like the one below, it is described that the file either needs to be included correctly or the database needs to be rebuilt:
https://hub.alfresco.com/t5/alfresco-content-services-forum/ensure-that-the-dir-root-property-c-alfres-4-alf-data-is/td-p/244923
But I have little influence on the creation of the data. What I don't understand, the properties file is created via the image, but is empty. On initial creation Alfresco works, after a restart no longer.

I have tried several things, but no matter how I configure alfresco, the problem is still there. I have tried compose, no proxy, bindings etc. The latest it fails on the next reboot. Except when I use compose with down & up. But then the restart command does not work either.

Therefore my request. How do they manage that the data remains existent after a restart of the container or the system? Is there a way to work around this error so that Alfresco can be restarted? In my attempts, I tested with the community version. For Docker stack, I adjusted the configuration a bit. If you have any questions, please do not hesitate to contact me.

Many greetings
Eric

gionn:
* Enable restarting the alfresco container without data loss

Now it's possible to run consecutive `compose up` without getting:

```
2025-02-26T14:08:49,603 [] ERROR [repo.admin.ConfigurationChecker] [main] CONTENT INTEGRITY ERROR: System content not found in content store: 'store://2025/2/26/13/49/a07593db-c44b-4edb-b593-dbc44bfedb5e.bin'
2025-02-26T14:08:49,603 [] ERROR [repo.admin.ConfigurationChecker] [main] Ensure that the 'dir.root' property './alf_data' is pointing to the correct data location.
2025-02-26T14:08:49,605 [] WARN  [context.support.XmlWebApplicationContext] [main] Exception encountered during context initialization - cancelling refresh attempt: org.alfresco.error.AlfrescoRuntimeException: 01260036 Ensure that the 'dir.root' property './alf_data' is pointing to the correct data location.
2025-02-26T14:08:49,953 [] ERROR [web.context.ContextLoader] [main] Context initialization failed
org.alfresco.error.AlfrescoRuntimeException: 01260036 Ensure that the 'dir.root' property './alf_data' is pointing to the correct data location.
	at org.alfresco.repo.admin.ConfigurationChecker.check(ConfigurationChecker.java:212) ~[alfresco-repository-23.4.1.1.jar:23.4.1.1]
	at org.alfresco.repo.admin.ConfigurationChecker.access$0(ConfigurationChecker.java:167) ~[alfresco-repository-23.4.1.1.jar:23.4.1.1]
	at org.alfresco.repo.admin.ConfigurationChecker$1$1.doWork(ConfigurationChecker.java:155) ~[alfresco-repository-23.4.1.1.jar:23.4.1.1]
```

Anonymous volumes still need to be cleaned up to recover space with `docker volume prune`.

Switching to anonymous volumes for sfs service for uniformity and because we don't really need a named tmpfs volume for it.

Fix #649

OPSEXP-3109

### DIFF
--- a/docker-compose/7.2.N-compose.yaml
+++ b/docker-compose/7.2.N-compose.yaml
@@ -145,7 +145,7 @@ services:
       retries: 3
       start_period: 10s
     volumes:
-      - shared-file-store-volume:/tmp/Alfresco
+      - /tmp/Alfresco
   share:
     image: quay.io/alfresco/alfresco-share:7.2.2.5
     mem_limit: 1g
--- a/docker-compose/7.2.N-compose.yaml
+++ b/docker-compose/7.2.N-compose.yaml
@@ -75,6 +75,8 @@ services:
       - "traefik.http.routers.alfrescomicrometer.rule=PathRegexp(`^/alfresco/(wc)?s(ervice)?/prometheus`)"
       - "traefik.http.middlewares.prometheusipfilter.ipallowlist.sourcerange=127.0.0.0/8"
       - "traefik.http.routers.alfrescomicrometer.middlewares=prometheusipfilter@docker"
+    volumes:
+      - /usr/local/tomcat/alf_data
   transform-router:
     mem_limit: 512m
     image: quay.io/alfresco/alfresco-transform-router:4.1.6
--- a/docker-compose/7.3.N-compose.yaml
+++ b/docker-compose/7.3.N-compose.yaml
@@ -138,7 +138,7 @@ services:
       retries: 3
       start_period: 10s
     volumes:
-      - shared-file-store-volume:/tmp/Alfresco
+      - /tmp/Alfresco
   share:
     image: quay.io/alfresco/alfresco-share:7.3.2.3
     mem_limit: 1g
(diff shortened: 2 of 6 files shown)


pmacius (approved):

alxgomz (commented):
If we rely on the local driver to dynamically create volumes, I see no good reason to keep using tmpfs (which anyway is documented to only work on linux) for sfs or even declare that volume at top level.
Can we align it with the other volumes?

PR #1334, opened 2025-04-01
OPSEXP-3189: add a release workflow for download trials

alxgomz:
Ref: OPSEXP-3189

### DIFF
--- a/.github/workflows/download-trials-release.yml
+++ b/.github/workflows/download-trials-release.yml
@@ -27,6 +27,11 @@ jobs:
           docker compose -f compose.yaml config -o docker-compose.yml && rm compose.yaml
           docker compose -f community-compose.yaml config -o community-docker-compose.yml && rm community-compose.yaml
 
+      - name: Set Commit Branch Name
+        id: set_commit_branch_name
+        run:
+          echo COMMIT_BRANCH_NAME=download-trial-release-${{ github.run_id }} >> $GITHUB_OUTPUT
+
       - name: Commit updated compose files
         uses: stefanzweifel/git-auto-commit-action@e348103e9026cc0eee72ae06630dbe30c8bf7a79  # v5.1.0
         with:
@@ -34,7 +39,7 @@ jobs:
             🛠 Update download trial compose files
           commit_user_name: ${{ vars.BOT_GITHUB_USERNAME }}
           commit_user_email: ${{ vars.BOT_GITHUB_EMAIL }}
-          branch: download-trial-release-${{ github.run_id }}
+          branch: ${{ steps.set_commit_branch_name.outputs.COMMIT_BRANCH_NAME }}
           create_branch: true
           file_pattern: >-
             docker-compose/docker-compose.yml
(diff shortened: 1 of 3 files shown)


.github/workflows/download-trials-release.yml
```diff
@@ -0,0 +1,56 @@
...
+
+on:
+  workflow_dispatch:
+
+env:
+  DEFAULT_BRANCH_NAME: master
+  TRIALS_BRANCH_NAME: download-trial
+
+jobs:
+  createPR:
```
pmacius, line 12: We use snake_case in other workflows
```suggestion
  create_pr:
```

.github/workflows/download-trials-release.yml
```diff
@@ -0,0 +1,56 @@
...
+          branch: ${{ steps.set_commit_branch_name.outputs.COMMIT_BRANCH_NAME }}
+          create_branch: true
+          file_pattern: >-
+            docker-compose/docker-compose.yml
+            docker-compose/community-docker-compose.yml
+
+      - name: Create Pull Request
+        env:
+          GH_TOKEN: ${{ secrets.BOT_GITHUB_TOKEN }}
+        run: >-
```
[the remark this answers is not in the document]
alxgomz: `needs` is used to carry stuff across different jobs. Here those are steps part of the same job so that's not needed

.github/workflows/download-trials-release.yml
```diff
@@ -0,0 +1,56 @@
...
+        with:
+          ref: ${{ env.TRIALS_BRANCH_NAME }}
+
+      - name: Generate Download trial compose files
+        working-directory: docker-compose
+        run: |
+          git fetch --no-tags --depth=1 origin ${{ env.DEFAULT_BRANCH_NAME }}:${{ env.DEFAULT_BRANCH_NAME }}
+          git restore --source ${{ env.DEFAULT_BRANCH_NAME }} compose.yaml community-compose.yaml commons/
+          docker compose -f compose.yaml config -o docker-compose.yml && rm compose.yaml
+          docker compose -f community-compose.yaml config -o community-docker-compose.yml && rm community-compose.yaml
```
gionn, line 28: rm is not really needed as you target specific file with the push action

pmacius (approved):

gionn (commented):
In the last test PR I see we are losing comments at the start of the file, not sure they are really worth keeping but need to check if that will break the existing integration which inject credentials

worst case pipe them in after dumping the compose config

alxgomz:
> In the last test PR I see we are losing comments at the start of the file, not sure they are really worth keeping but need to check if that will break the existing integration which inject credentials
> 
> worst case pipe them in after dumping the compose config

I did check that and it seems ok: https://github.com/AlfrescoEAD/service-download.alfresco.com/blob/88bcd040f0d744ceb79d1898552f34f890191a81/downloads/downloads_app/compose_marketo_injector.py#L54

gionn (approved):

PR #1356, opened 2025-05-07
OPSEXP-3134 Add an HXInsight docker compose override file

pmacius:
Ref: OPSEXP-3134

### DIFF
--- a/docker-compose/hxi-overrides.yaml
+++ b/docker-compose/hxi-overrides.yaml
@@ -1,0 +1,58 @@
+services:
+  alfresco:
+    image: localhost/alfresco/alfresco-content-repository:latest
+    environment:
+      JAVA_OPTS: >-
+        -Ddb.driver=org.postgresql.Driver
+        -Ddb.username=alfresco
+        -Ddb.password=alfresco
+        -Ddb.url=jdbc:postgresql://postgres:5432/alfresco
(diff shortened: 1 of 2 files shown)


docker-compose/hxi-overrides.yaml
```diff
@@ -0,0 +1,58 @@
...
+        -Daos.baseUrlOverwrite=http://localhost:8080/alfresco/aos
+        -Dmessaging.broker.url="failover:(nio://activemq:61616)?timeout=3000&jms.useCompression=true"
+        -Ddeployment.method=DOCKER_COMPOSE
+        -Dtransform.service.enabled=true
+        -Dtransform.service.url=http://transform-router:8095
+        -Dsfs.url=http://shared-file-store:8099/
+        -DlocalTransform.core-aio.url=http://transform-core-aio:8090/
+        -Ddsync.service.uris=http://localhost:9090/alfresco
+        -XX:MinRAMPercentage=50
+        -XX:MaxRAMPercentage=80
```
alxgomz, line 28: It bothers me we have to maintain this properties here too. They do not get updated very frequently so I bet we would forget about it quite easily. On the other hand I couldn't find a way to concatenate strings in compose files so the only way out would be to have a shell wrapper that does the processing for us...

alxgomz (approved):

PR #1360, opened 2025-05-24
💅 Fail template on missing 'sharedSecret' when Solr is enabled & uses "secret" to secure communication

deplab:
This PR updates `community_values.yaml` to include a note about the `global.search.sharedSecret`, and that there are three modes of communication, with `secret` mode of communication being the default.

While deploying ACM to EKS, I noticed some pods were crashing when using the `community_values.yaml`. After some investigation, I found that the issue was due to the `sharedSecret` being set to `null`, however, this option was not available in the `community_values.yaml`, and it took some time to find out that this existed - especially for someone trying Alfresco ACM the first time. Hopefully, adding this information to the `community_values.yaml` will save some time for others 😊

### DIFF
--- a/.github/workflows/helm-community.yml
+++ b/.github/workflows/helm-community.yml
@@ -50,7 +50,7 @@ jobs:
     needs:
       - build_vars
     env:
-      REGISTRY_SECRET_NAME: ${{ ! github.event.repository.fork && 'regcred' || '' }}
+      REGISTRY_SECRET_NAME: ${{ github.event.pull_request.head.repo.fork && '' || 'regcred' }}
     strategy:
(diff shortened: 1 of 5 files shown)


gionn:
Hello @deplab, does reading https://alfresco.github.io/acs-deployment/docs/helm/desktop-deployment.html#community-localhost-deployment would have helped?

For sure there is plenty of room to improve the helm docs as some information is scattered across the different deployment examples which may be worth to unify

alxgomz (approved):

alxgomz:
LGTM. I took the liberty to add tests (which I forgot to ask in the first place) and a build fix

PR #1390, opened 2025-09-05
OPSEXP-3492 Replace bitnami postgres with our chart

pmacius:
ref: OPSEXP-3485 OPSEXP-3492


helm/alfresco-content-services/23.N_values.yaml
```diff
@@ -65,10 +65,10 @@ alfresco-control-center:
     tag: 9.3.0
 postgresql:
   image:
-    tag: 15.10.0
+    tag: 15.10
 postgresql-sync:
   image:
-    tag: 14.10.0
+    tag: 14.10
```
alxgomz, line 71: I think it's safe to align with ACS version. the dsync CI even test again a latest tag of postgres

PR #1400, opened 2025-09-22
OPSEXP-3458 Add example with telegraf

pmacius:
ref: OPSEXP-3458

### DIFF
--- a/docs/docker-compose/examples/with-telegraf-metrics.md
+++ b/docs/docker-compose/examples/with-telegraf-metrics.md
@@ -18,7 +18,7 @@ This example uses config files from `docs/docker-compose/examples/config` folder
 - telegraf-overrides.yaml
 - telegraf.conf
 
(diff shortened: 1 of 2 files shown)


docs/docker-compose/examples/with-telegraf-metrics.md
```diff
@@ -0,0 +1,143 @@
...
+```toml
+[[outputs.influxdb_v2]]
+  bucket = "alfresco"
+  organization = "alfresco"
+  timeout = "5s"
+  token = "influx"
+  urls = [
+    "http://influxdb2:8086"
+  ]
+```
```
gionn, line 84: not sure if make sense to document influx specifically here, but anyway telegraf is part of the same ecosystem so it's a natural choice. we should probably specify better in the introduction that we are targeting influxdb but we can target any supported output (link to the outputs?)

PR #1410, opened 2025-10-08
OPSEXP-3321: remove 7.3 support from CI

ffjdm:


pmacius (approved):
Im cool with removing now but support page mentions November https://www.hyland.com/en/resources/alfresco-product-support-status. Lets wait for @gionn @alxgomz

PR #1412, opened 2025-10-14
OPSEXP-3537 Bump tags for Virgo release

siddavamshi4:
Ref: OPSEXP-3537

### DIFF
--- a/docs/helm/upgrades.md
+++ b/docs/helm/upgrades.md
@@ -16,7 +16,7 @@ Releases](https://github.com/Alfresco/acs-deployment/releases).
 Here follows a more detailed explanation of any breaking change grouped by
 version in which they have been released.
 
-## Unreleased
+## 10.0.0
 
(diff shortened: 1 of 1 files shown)


helm/acs-sso-example/Chart.yaml
```diff
@@ -20,7 +20,7 @@ description: |
   :warning: All components have persistence disabled so all data is lost after a
   deployment is destroyed or rolled back!
 type: application
-version: 1.4.0
+version: 1.5.0
```
pmacius: No changes in this chart so I think no bump

pmacius (commented):
Change https://github.com/Alfresco/acs-deployment/blob/8a3f51376f72340c0b940acf6a9407a65f02f0e4/docs/helm/upgrades.md#unreleased to 10.0.0

pmacius (approved):

PR #1428, opened 2025-12-16
OPSEXP-3702: Support ActiveMQ 6.2 in pre-release with credentials

siddavamshi4:
Ref: OPSEXP-3702
This PR adds support for Activemq 6.2 in the pre-release Docker compose and Helm charts.
It also configures ActiveMQ broker authentication across multiple services( as we have enabled jaas authentication on activemq)


alxgomz (approved):

siddavamshi4:
> let's wait for the revert of default authentication on 5.x
As authentication is reverted on 5.x, we can proceed with the PR

PR #1540, opened 2026-07-20
OPSEXP-4174 Add Elasticsearch batch indexing option to ACS Community

gionn:
Wires the new `alfresco-search-community` chart (Elasticsearch batch indexing, `docker.io/alfresco/alfresco-elasticsearch-batch-indexing`) into the `alfresco-content-services` umbrella so the Community deployment can run Elasticsearch-based search on the Kubernetes clusters we manage, as an alternative to Solr.

## Behaviour

Solr stays the default for a plain community install (`helm install ... -f community_values.yaml`), so existing installs are unaffected. Elasticsearch batch indexing is opt-in via `alfresco-search-community.enabled=true`, which also requires `alfresco-search.enabled=false`, `elasticsearch.enabled=true` and `alfresco-repository.configuration.search.flavor=elasticsearch`.

The `search.flavor` helper maps the community chart to the `elasticsearch` flavor and fails fast on invalid combinations: Solr + community both enabled, `alfresco-search-enterprise` + community both enabled (note enterprise search defaults to enabled, so a community install must disable it — `community_values.yaml` does), a repository search flavor other than `elasticsearch`, or a missing Elasticsearch backend.

## Wiring

The batch indexer reads everything from the existing single source of truth — no new keys in `config-infrastructure.yaml`:

- `SEARCH_URL`, `REPOSITORY_URL`, `DATABASE_URL` from the shared `alfresco-infrastructure` ConfigMap
- DB credentials from `alfresco-cs-database`, ES credentials (empty, no-auth ES) from `alfresco-search-secret`

The accepted-content-media-types cache is disabled (`ALFRESCO_ACCEPTEDCONTENTMEDIATYPESCACHE_ENABLED=false`), per engineering guidance, so no aggregating transform endpoint is required — Kubernetes community has no `transform-core-aio` all-in-one and the transform router is enterprise-only. The chart still mandates a `transform.url`/`sharedSecret.value`, so both are set to inert placeholders.

## Repository image requirement

The Community repository's `elasticsearch` search subsystem is only available from `alfresco/alfresco-content-repository-community:26.2.0-A.23`; the default `community_values.yaml` tag (`26.1.0`) lacks it and fails bootstrap with `No bean named 'elasticsearch' available`. The Elasticsearch integration-test overlay therefore pins the repository image to `26.2.0-A.23`.

## Tests / CI

- Added five cases to `tests/search_test.yaml` (helm-unittest) covering the umbrella-owned logic: community-enabled ⇒ `elasticsearch` flavor + `SEARCH_URL`, the Solr/community and enterprise/community mutual-exclusion fails, the flavor-mismatch fail, and the missing-Elasticsearch-backend fail. Full chart suite passes 36/36. Subchart-owned Deployment/env wiring is intentionally not re-tested here.
- `helm-community.yml` now installs both the Solr path (unchanged) and the Elasticsearch path (`community_values.yaml` + new `test/community-es-integration-test-values.yaml`) so CI validates both on KIND, with the two `alfresco-content-services` matrix jobs labelled `(solr)` / `(elasticsearch)` for readability. The workflow also gained an explicit least-privilege `permissions: contents: read` block.

## Notes

- Depends on `alfresco-search-community` `0.1.0-alpha.0` (pinned; expect to bump on the next pre-release/GA).
- One runtime item to confirm on the KIND run: whether `REPOSITORY_URL` (`http://…:80/alfresco`, with context path) is accepted as `ALFRESCO_ACS_URL`; if not, override `alfresco-search-community.repository.url` with a base URL.

OPSEXP-4174

### DIFF
--- a/test/community-es-integration-test-values.yaml
+++ b/test/community-es-integration-test-values.yaml
@@ -29,6 +29,8 @@ elasticsearch:
     readinessProbe:
       initialDelaySeconds: 100
 alfresco-repository:
+  image:
+    tag: 26.2.0-A.23
   configuration:
     search:
       flavor: elasticsearch
--- a/test/community-es-integration-test-values.yaml
+++ b/test/community-es-integration-test-values.yaml
@@ -1,0 +1,146 @@
+---
+# Overlay on top of community_values.yaml that switches the community stack from
+# Solr to the Elasticsearch batch indexing component (alfresco-search-community),
+# with resource sizing tuned for the KIND integration runner.
+alfresco-search:
+  enabled: false
+alfresco-search-community:
+  enabled: true
+  resources:
+    requests:
+      cpu: "0.01"
+      memory: "256Mi"
+    limits:
+      cpu: "2"
+      memory: "1Gi"
+  environment:
+    JAVA_OPTS: "-Xms256m -Xmx768m"
+elasticsearch:
+  enabled: true
+  elasticsearch:
+    resources:
+      requests:
+        cpu: "0.01"
+        memory: "512Mi"
+      limits:
+        memory: "1.5Gi"
+    livenessProbe:
+      initialDelaySeconds: 100
+    readinessProbe:
+      initialDelaySeconds: 100
+alfresco-repository:
+  configuration:
+    search:
+      flavor: elasticsearch
+  resources:
+    requests:
+      cpu: "0.01"
+      memory: "1Gi"
+    limits:
+      cpu: "2"
+      memory: "2Gi"
+  environment:
+    JAVA_OPTS: >-
+      -XX:MaxRAMPercentage=65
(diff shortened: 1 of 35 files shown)


test/community-es-integration-test-values.yaml
```diff
@@ -0,0 +1,148 @@
...
+        memory: "512Mi"
+      limits:
+        memory: "1.5Gi"
+    livenessProbe:
+      initialDelaySeconds: 100
+    readinessProbe:
+      initialDelaySeconds: 100
+alfresco-repository:
+  image:
+    tag: 26.2.0-A.23
```
pmacius, line 33: We need to remember about this when releasing

pmacius (approved):

PR #1575, opened 2026-08-19
OPSEXP-4240 Fix intermittent Newman failures in Enterprise Helm CI

siddavamshi4:
OPSEXP-4240
## Summary

Improve the reliability of ACS Enterprise Helm CI tests by running the Search Enterprise reindexing job as a post-install hook, after the ACS resources and Elasticsearch deployment have been installed, before Newman tests execute.

#### Behavior change:

hookExecution: false: reindexing runs as a normal Kubernetes Job during installation; Helm does not treat it as a lifecycle hook.
hookExecution: post-install: reindexing runs as a Helm post-install hook; helm install --wait waits for it to finish.

### DIFF
--- a/.github/workflows/helm-enterprise.yml
+++ b/.github/workflows/helm-enterprise.yml
@@ -136,7 +136,7 @@ jobs:
           --values helm/${{ matrix.name }}/${{ matrix.values }}
           --values test/enterprise-integration-test-values.yaml
           ${{ steps.configurable-extra-values.outputs.helm_install_params }}
-          --wait --timeout ${{ env.HELM_INSTALL_TIMEOUT }}
+          --timeout ${{ env.HELM_INSTALL_TIMEOUT }}
 
       - name: Wait for deployments to be ready
         uses: Alfresco/alfresco-build-tools/.github/actions/kubectl-wait@v18.24.1
--- a/.github/workflows/helm-enterprise.yml
+++ b/.github/workflows/helm-enterprise.yml
@@ -67,7 +67,7 @@ jobs:
       REGISTRY_SECRET_NAME: regcred
       POD_METRICS_LOG: /tmp/pod-metrics.log
       NODE_METRICS_LOG: /tmp/node-metrics.log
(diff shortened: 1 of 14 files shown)


.github/workflows/helm-enterprise.yml
```diff
@@ -135,6 +136,7 @@ jobs:
           --values helm/${{ matrix.name }}/${{ matrix.values }}
           --values test/enterprise-integration-test-values.yaml
           ${{ steps.configurable-extra-values.outputs.helm_install_params }}
+          --wait --timeout ${{ env.HELM_INSTALL_TIMEOUT }}
```
gionn, line 139: why the explicit `--wait`? all the next wait jobs are there to get a better failure than "operation timeout" which helm gives - if you add --wait there the next ones are no-op

gionn: and anyway the tests still fail

.github/workflows/helm-enterprise.yml
```diff
@@ -135,6 +136,7 @@ jobs:
...
 
       - name: Wait for deployments to be ready
```
gionn: those wait steps are not doing anything helpful now

siddavamshi4: without wait it fails
```
Deleting outdated charts
Error: INSTALLATION FAILED: failed post-install: 1 error occurred:
	* timed out waiting for the condition
```	
https://github.com/Alfresco/acs-deployment/actions/runs/32978812097/job/98210133590#step:11:35

gionn: I didn't suggest to remove the --wait on the helm install, the suggestion in on L141 and I was referring to the 3 different Wait* steps after the main helm install

gionn (changes requested):
remove all the kubectl-wait steps

gionn (approved):

pmacius (approved):

alxgomz (approved):

